PR #10, opened 2017-11-09
Running test in a single folder

ibr-jabrego:
All tests will now run on a single folder, to prevent generating too many RDFs or NonRDFs in the root folder.


awoods:
@sprater Please merge this PR if it works on your side.

PR #12, opened 2017-11-29
Fedora Test Suite Authentication Version

ibr-jabrego:
Fedora Test Suite uses new parameters when running. -h is required, -u -p not required if server does not have authentication.

New tests have been added.

### DIFF
--- a/src/main/java/com/ibr/fedora/App.java
+++ b/src/main/java/com/ibr/fedora/App.java
@@ -20,15 +20,22 @@
  */
 package com.ibr.fedora;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-
+import org.apache.commons.cli.BasicParser;
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.CommandLineParser;
+import org.apache.commons.cli.HelpFormatter;
+import org.apache.commons.cli.Option;
+import org.apache.commons.cli.Options;
+import org.apache.commons.cli.ParseException;
 import org.testng.TestNG;
 import org.testng.xml.XmlClass;
 import org.testng.xml.XmlSuite;
 import org.testng.xml.XmlTest;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 public class App {
 private App() { }
@@ -39,6 +46,37 @@ private App() { }
  * @param args
  */
 public static void main(final String[] args) {
+final Options options = new Options();
+
+final Option host = new Option("h", "host", true, "host address");
+host .setRequired(true);
+options.addOption(host);
+
+final Option user = new Option("u", "user", true, "user login");
+user.setRequired(false);
+options.addOption(user);
+
+final Option password = new Option("p", "password", true, "user's password");
+password.setRequired(false);
+options.addOption(password);
+
+final CommandLineParser parser = new BasicParser();
+final HelpFormatter formatter = new HelpFormatter();
+CommandLine cmd;
+
+try {
+cmd = parser.parse(options, args);
+} catch (ParseException e) {
+System.out.println(e.getMessage());
+formatter.printHelp("Fedora Test Suite", options);
+System.exit(1);
+return;
+}
+
+String inputHost = cmd.getOptionValue("host");
+final String inputUser = cmd.getOptionValue("user") == null ? "" : cmd.getOptionValue("user");
+final String inputPassword = cmd.getOptionValue("password") == null ? "" : cmd.getOptionValue("password");
+
 final TestNG testng = new TestNG();
 final XmlSuite suite = new XmlSuite();
 suite.setName("ldptest");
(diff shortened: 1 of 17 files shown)


src/main/java/com/ibr/fedora/App.java
```diff
@@ -39,6 +46,37 @@ private App() { }
...
+
+final Option password = new Option("p", "password", true, "user's password");
+password.setRequired(false);
+options.addOption(password);
+
+final CommandLineParser parser = new BasicParser();
+final HelpFormatter formatter = new HelpFormatter();
+CommandLine cmd;
+
+try {
```
awoods: The left-justification in the class is a little surprising.
Is there a reason for not using a more usual indentation scheme?

Note to self: The checkstyle should be configured to flag this.

src/main/java/com/ibr/fedora/testsuite/Container.java
```diff
@@ -0,0 +1,185 @@
...
+
+/**
+ * Authentication
+ * @param username
+ * @param password
+ */
+@BeforeClass
+@Parameters({"param2", "param3"})
+public void auth(final String username, final String password) {
+this.username = username;
```
awoods: Same note as before on formatting.

src/main/java/com/ibr/fedora/testsuite/ExternalBinaryContent.java
```diff
@@ -20,37 +20,55 @@
...
+this.username = username;
+this.password = password;
+}
+
     /**
+     * 3.8-A
      * @param host
      */
-    @Test(priority = 19)
+    @Test(priority = 34)
```
awoods: Ideally, we would not need to enforce an order to the tests. Each test should create its own scaffolding.
...something to keep in mind for a future refactoring.

src/main/java/com/ibr/fedora/testsuite/HttpDelete.java
```diff
@@ -25,81 +25,77 @@
...
-            .delete(resource)
-            .then()
-            .log().all()
-            .statusCode(204);
-
-    ps.append("\n -Case End- \n").close();
+@BeforeClass
+@Parameters({"param2", "param3"})
+public void auth(final String username, final String password) {
+this.username = username;
```
awoods: Why no indentation :(

src/main/java/com/ibr/fedora/testsuite/HttpPut.java
```diff
@@ -24,32 +24,54 @@
...
 @Parameters({"param1"})
 public void httpPut(final String host) throws FileNotFoundException {
     final PrintStream ps = TestSuiteGlobals.logFile();
 
-    ps.append("\n8." + tl.httpPut()[0] + "-" + tl.httpPut()[1]).append("\n");
+    ps.append("\n20." + tl.httpPut()[0] + "-" + tl.httpPut()[1]).append("\n");
     ps.append("Request:\n");
     final String resource =
             RestAssured.given()
+.auth().basic(this.username, this.password)
```
awoods: Ideally, we would not have to include `.auth()` in so many places. In other words, it would be good to refactor the request construction into a generalized method.
...but can be restructured later.

src/main/java/com/ibr/fedora/testsuite/HttpPut.java
```diff
@@ -65,23 +87,184 @@ public void httpPut(final String host) throws FileNotFoundException {
 }
 
 /**
+ * 3.4.1-A
+ * @param host
+ * @throws FileNotFoundException
+ */
+@Test(priority = 21)
+@Parameters({"param1"})
+public void updateTriples(final String host) throws FileNotFoundException {
```
awoods: formatting :(

src/main/java/com/ibr/fedora/testsuite/Ldpnr.java
```diff
@@ -0,0 +1,149 @@
...
+public void auth(final String username, final String password) {
+this.username = username;
+this.password = password;
+}
+
+/**
+ * 3.1.1.-A
+ * @param host
+ */
+@Test(priority = 1)
```
awoods: formatting :(

awoods:
Unless there is a reason for the left-justified formatting, I would prefer that this PR fix that issue before being merged.
The other topics:
- test order
- refactoring `auth()` etc

...can wait until later.

src/main/java/com/ibr/fedora/App.java
ibr-jabrego: Currently, we have not added indented the lines because of the check style. If we indent the lines we get an error. The error is: Line contains a tab character, when we indent it. Should we indent the code and ignore the check style.

<error line="64" column="1" severity="error" message="Line contains a tab character." source="com.puppycrawl.tools.checkstyle.checks.whitespace.FileTabCharacterCheck"/>

src/main/java/com/ibr/fedora/testsuite/ExternalBinaryContent.java
ibr-jabrego: We are looking for a solution that will use the order in which classes are added to the test to TestNG as the test order. If the priority is not included the test results are order and grouped by status, FAIL first, PASS second. We will continue to look for the solution.

src/main/java/com/ibr/fedora/App.java
awoods: Agreed. Tabs are not allowed, but spaces are:
https://wiki.duraspace.org/display/FF/Code+Style+Guide

src/main/java/com/ibr/fedora/testsuite/ExternalBinaryContent.java
awoods: Thanks. It seems like a maintenance burden to have to be updating lots of `priority` values when new tests are added... but I also see the benefit in being able to compare two test results side-by-side, which would be difficult with different orderings.

It would be nice if this could be handled on the reporting side, and not in the tests... but this is not a critical issue.

src/main/java/com/ibr/fedora/App.java
ibr-jabrego: I understand, no problem, we will make the indentation changes ASAP.

src/main/java/com/ibr/fedora/testsuite/ExternalBinaryContent.java
sprater: But I agree with Andrew:  not a high priority right now.

PR #37, opened 2018-02-06
Added missing header for POST LDPC

ibr-jabrego:
Added missing headers to tests that used POST create LDPC.

Related to:
https://github.com/fcrepo4-labs/Fedora-API-Test-Suite/issues/35


src/main/java/com/ibr/fedora/TestSuiteGlobals.java
```diff
@@ -60,7 +65,11 @@ public static String containerTestSuite(final String baseurl, final String user,
         final Response res = RestAssured.given()
         .auth().basic(user, pass)
```
sprater: fix indentation

ibr-jabrego: Thank you, I will fix the indentation.

PR #60, opened 2018-04-25
Renames packages to align with other fedora ecosystem projects.

dbernstein:
The base package name was renamed to org.fcrepo.spec.testsuite. The testsuite directory was renamed to tests.  No functional changes.


src/main/java/com/ibr/fedora/report/EarlCoreReporter.java
```diff
@@ -1,99 +0,0 @@
-/**
```
awoods: This class should likely not be deleted.

src/main/java/com/ibr/fedora/report/EarlReporter.java
```diff
@@ -1,149 +0,0 @@
...
-import org.apache.jena.vocabulary.DCTerms;
-import org.apache.jena.vocabulary.RDF;
-import org.testng.IReporter;
-import org.testng.IResultMap;
-import org.testng.ISuite;
-import org.testng.ISuiteResult;
-import org.testng.ITestContext;
-import org.testng.xml.XmlSuite;
-
-public class EarlReporter extends EarlCoreReporter implements IReporter {
```
awoods: This class should likely not be deleted.

src/main/java/com/ibr/fedora/report/HtmlReporter.java
```diff
@@ -1,196 +0,0 @@
...
-import org.rendersnake.StringResource;
-import org.testng.IReporter;
-import org.testng.IResultMap;
-import org.testng.ISuite;
-import org.testng.ISuiteResult;
-import org.testng.ITestContext;
-import org.testng.ITestResult;
-import org.testng.xml.XmlSuite;
-
-public class HtmlReporter implements IReporter {
```
awoods: This class should likely not be deleted.

src/main/java/org/fcrepo/spec/testsuite/TestSuiteGlobals.java
```diff
@@ -41,7 +41,7 @@
 public abstract class TestSuiteGlobals {
     public static String cssReport = "reportStyle.css";
     public static String outputDirectory = "report";
-    public static String outputName = "testsuite";
```
awoods: Is there a reason for changing the output report name?

src/main/java/org/fcrepo/spec/testsuite/test/Container.java
```diff
@@ -18,13 +18,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.ibr.fedora.testsuite;
+package org.fcrepo.spec.testsuite.test;
```
awoods: What is the purpose of the additional `test` package?

dbernstein: The root path was com.ibr.fedora.  All the tests were in  com.ibr.fedora.testsuite
Thus com.ibr.fedora became org.fcrepo.spec.testsuite and com.ibr.fedora.testsuite became org.fcrepo.spec.testsuite.test  since the since the "testsuite" dir contains all project files whereas the "test" directory is the location of all the tests.

src/main/java/org/fcrepo/spec/testsuite/TestSuiteGlobals.java
dbernstein: I think this must have been updated automatically when I changed the artifact name from testSuite to testsuite.  I can change this back.

dbernstein: I will change this back.

.gitignore
```diff
@@ -18,4 +18,3 @@ maven-eclipse.xml
 atlassian-ide-plugin.xml
 test-output
 /bin
-report/
```
awoods: I believe we still want to ignore the generated `report/` dir.

dbernstein: yes. you're right. will fix.

PR #79, opened 2018-05-02
Ensures that test class and method name are written to the  testing log.

dbernstein:
Also fixes a related bug (output log missing after testsuite completes),
improves the layout of the testing log, and removes related cruft.
Resolves https://github.com/fcrepo4-labs/Fedora-API-Test-Suite/issues/66


awoods:
I think user-facing text should end up  in the HTML report... such as the TestInfo.getDescription().
The Class and method of tests that is being printed to the log file is likely sufficient for https://github.com/fcrepo4-labs/Fedora-API-Test-Suite/issues/66

dbernstein:
@awoods:  good point - see my changes.

awoods:
@dbernstein : Looks like we have a merge conflict to be resolved before merging.

PR #80, opened 2018-05-02
Outputs requirement level in result report.

dbernstein:
Resolves:  https://github.com/fcrepo4-labs/Fedora-API-Test-Suite/issues/78


dbernstein:
@awoods :  I just pushed a commit to address the issue of FAIL text not being red.

PR #110, opened 2018-06-26
Distinct name for each test

### LINKED ISSUE #107: Split naming of tests currently called "3.9-A" (3), "3.9-C" (2), and "3.9-J-2" (2)
The test execution report does not sensibly report results for multiple tests with the same name. The remaining cases seem to be found with:

```
Fedora-API-Test-Suite (master)> grep -r 'setupTest("' src | perl -pe 's/.*setupTest\("([^"]+)".*/$1/' | sort | uniq -c | sort | tail -4
   1 3.9-I-2
   2 3.9-C
   2 3.9-J-2
   3 3.9-A
```

all in https://github.com/fcrepo4-labs/Fedora-API-Test-Suite/blob/master/src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java

zimeon:
Fixes #107

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
@@ -58,14 +58,14 @@ public class ExternalBinaryContent extends AbstractTest {
     }
 
     /**
-     * 3.9-A PostCreate
+     * 3.9-A-1 PostCreate
      *
      * @param uri
      */
     @Test(groups = {"SHOULD"})
     @Parameters({"param1"})
     public void postCreateExternalBinaryContent(final String uri) throws FileNotFoundException {
-        final TestInfo info = setupTest("3.9-A", "postCreateExternalBinaryContent",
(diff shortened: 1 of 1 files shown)


src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -869,7 +869,7 @@ public void respondWantDigestTwoSupportedQvalueNonZeroExternalBinaryContentHead(
     public void respondWantDigestNonSupportedExternalBinaryContent(final String uri)
         throws FileNotFoundException {
         final TestInfo info =
-            setupTest("3.9-J-2", "respondWantDigestNonSupportedExternalBinaryContent",
+            setupTest("3.9-J-1", "respondWantDigestNonSupportedExternalBinaryContent",
```
awoods: Instead of `3.9-J-1`, should this instead be named `3.9.3`?
https://fedora.info/2018/06/25/spec/#redirect-and-proxy

zimeon:
This is expanding the scope a bit but I agree 3.9-J-1/2 are actually for 3.9.3 so have renamed as 3.9.3-A/B

Issue #116, opened 2018-06-27
Ensure setup requests succeed


awoods:
Currently, the setup requests, such as:
https://github.com/fcrepo4-labs/Fedora-API-Test-Suite/blob/master/src/main/java/org/fcrepo/spec/testsuite/crud/HttpPut.java#L68

...do not verify that the target resource was created/updated. 
If setup requests fail, the test should preemptively fail.

PR #118, opened 2018-06-28
Ensure HTTP requests have expected responses

awoods:
Resolves: #116


awoods:
Note: Please squash/merge this PR... the two commits should be combined in the final history.

PR #124, opened 2018-06-28
Fix test 3.6.1-C

### LINKED ISSUE #123: Test 3.6.1-C - Test assumes particular serialization of explanation response
This test for the string "ldp#contains":

https://github.com/fcrepo4-labs/Fedora-API-Test-Suite/blob/33d846992842eacbfcb36727b09efbaa166e3304/src/main/java/org/fcrepo/spec/testsuite/crud/HttpPut.java#L166

will fail if the explanation uses a serialization that uses a prefix to write the predicate as `ldp:contains` or `ns1:contains` -- both of which would meet the specification requirement.

I think the tests should just check for match on `contains`, but might also add a match on `some-name`.

zimeon:
Fixes #123

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/crud/HttpPut.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/crud/HttpPut.java
@@ -157,13 +157,14 @@ public class HttpPut extends AbstractTest {
 
         final String body2 = doGet(locationHeader).asString();
 
-        final String newBody = body2.replace("containedFolderSlug", "some-name");
+        final String newBody = body2.replace("containedFolderSlug", "conflicting-contained-resource");
 
         ps.append("PUT Request: \n");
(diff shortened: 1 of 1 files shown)


src/main/java/org/fcrepo/spec/testsuite/crud/HttpPut.java
```diff
@@ -157,13 +157,14 @@ public void httpPutUpdateDisallowedTriplesResponse(final String uri) {
 
         final String body2 = doGet(locationHeader).asString();
 
-        final String newBody = body2.replace("containedFolderSlug", "some-name");
+        final String newBody = body2.replace("containedFolderSlug", "conflicting-contained-resource");
```
awoods: I am inclined to replace many of these `slug`s or hardcoded resource names with UUIDs, but this is a constructive step forward.

Issue #131, opened 2018-07-05
Should a server be expected to send an ETag in response to POST/201?


zimeon:
The change in #130 to address #88 assumes that an ETag is sent in response to the POST request that creates a resource:

https://github.com/awoods/Fedora-API-Test-Suite/blob/bf2dfbf8a8b521a97228098824ab928a8dcb7f15/src/main/java/org/fcrepo/spec/testsuite/crud/HttpPut.java#L98-L100

I'm not sure it is reasonable to expect an ETag in anything other than response to GET/HEAD per https://tools.ietf.org/html/rfc7232#section-2.4 ?

Issue #88, opened 2018-06-20
Test 3.6.X -- missing If-Match header in request


zimeon:
Per [LDP 4.2.4.5](https://www.w3.org/TR/ldp/#ldpr-put-precond), servers _SHOULD_ require `If-Match` when implementing PUT and, if they do, _MUST_ respond with 428 if it isn't present in a request. It seems that the current implementation of the 3.6.X tests in https://github.com/fcrepo4-labs/Fedora-API-Test-Suite/blob/c30f47aa55c89ed1336cf4944aff104ce5e6bfa5/src/main/java/org/fcrepo/spec/testsuite/crud/HttpPut.java#L89-L236 do not send an `If-Match` then incorrectly fail if they get a 428 back. I think they should send `If-Match` as that is harmless if sent to a server that does not implement it, needed if a server does.

PR #132, opened 2018-07-05
Get ETag from GET not POST

awoods:
Related to: #88
Resolves: #131


zimeon:
This PR looks good wrt the issue of getting the ETag with GET rather than expecting from POST.

However, it seems that this code is sending back the ETag without the double-quotes, so if the GET response includes:

```
Etag: "b866f522a1ef62d6d5df39ea69780c5f"
```
then the following request has:

```
If-Match: b866f522a1ef62d6d5df39ea69780c5f
```
where the double are actually an integral part of the tag value, see  https://tools.ietf.org/html/rfc7232#section-2.3 and https://tools.ietf.org/html/rfc7232#section-3.1

Issue #137, opened 2018-07-24
Test 4.1.1-A-1 and other versioning tests - Cannot rely on headers from POST the creates resource


zimeon:
Similar to #131, I don't think tests can rely on the response headers from the POST request that creates a resource to include the headers one expects from GET. The first instance I see of this is in [4.1.1-A-1](https://github.com/fcrepo/Fedora-API-Test-Suite/blob/0ce06cef701bf6cc6261d4d3438022fed64742d2/src/main/java/org/fcrepo/spec/testsuite/versioning/LdprvHttpGet.java#L69-L73) where I think the test requires a `timegate` link in the response to POST. I think the test should instead just rely on 201/`Location` from the POST and then do a GET on that location. I think there are other cases in the 4.1.X tests too.

PR #138, opened 2018-07-27
Test 4.1.1-A-1: Collect resource headers from GET, not initial POST

awoods:
Resolves: https://github.com/fcrepo/Fedora-API-Test-Suite/issues/137


zimeon:
Looks good but does not address any of the other 4.1.X tests -- I think some of those share the problem (e.g. 4.1.1-A-2)

Issue #136, opened 2018-07-24
Test 3.7-F -- Passes only if successful PATCH gives 204, spec requires only 2xx


zimeon:
The [`doPatch`](https://github.com/fcrepo/Fedora-API-Test-Suite/blob/6eae7c6d87f8d63096612899ee01047611aad924/src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java#L292-L298) method has a built-in test for 204 response. Given the [spec](https://fedora.info/2018/06/25/spec/#http-patch):

> A successful PATCH request must respond with a 2xx status code; the specific code in the 2xx range may vary according to the response body or request state.

and similarly in [RFC5789](https://tools.ietf.org/html/rfc5789#section-2.1):

> The 204 response code is used because the response does not carry a
   message body (which a response with the 200 code would have).  Note
   that other success codes could be used as well.

I think the test should be just for 2xx not 204.

PR #139, opened 2018-07-27
Use status ranges when verifying responses

awoods:
Resolves: https://github.com/fcrepo/Fedora-API-Test-Suite/issues/136


zimeon:
:+1: looks good

Issue #141, opened 2018-08-03
Work Test 4.3.6: An implementation that does support DELETE should do so by both removing the LDPCv and removing the versioning interaction model from the original LDPRv.


dbernstein:
LdpcvHttpDelete.ldpcvThatAdvertisesDeleteShouldRemoveContainerAndMementos()

dbernstein:
https://github.com/fcrepo/Fedora-API-Test-Suite/pull/142

PR #142, opened 2018-08-03
Adds test for 4.3.6.

dbernstein:
Resolves: https://github.com/fcrepo/Fedora-API-Test-Suite/issues/141

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpDelete.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpDelete.java
@@ -17,8 +17,13 @@
  */
 package org.fcrepo.spec.testsuite.versioning;
 
+import static org.fcrepo.spec.testsuite.Constants.ORIGINAL_RESOURCE_LINK_HEADER;
+import static org.testng.AssertJUnit.assertEquals;
+
+import io.restassured.response.Response;
 import org.fcrepo.spec.testsuite.TestInfo;
 import org.testng.annotations.Parameters;
+import org.testng.annotations.Test;
 
 /**
  * Tests for DELETE requests on LDP Version Container Resources
@@ -39,11 +44,11 @@ public class LdpcvHttpDelete extends AbstractVersioningTest {
     }
 
     /**
-     * 4.3.6
+     * Tests section 4.3.6
      *
      * @param uri The repository root URI
      */
-    //@Test(groups = {"SHOULD"})
+    @Test(groups = {"SHOULD"})
     @Parameters({"param1"})
     public void ldpcvThatAdvertisesDeleteShouldRemoveContainerAndMementos(final String uri) {
         final TestInfo info = setupTest("4.3.6", "ldpcvThatAdvertisesDeleteShouldRemoveContainerAndMementos",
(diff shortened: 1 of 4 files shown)


src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpDelete.java
```diff
@@ -53,6 +58,25 @@ public void ldpcvThatAdvertisesDeleteShouldRemoveContainerAndMementos(final Stri
...
+        if (hasHeaderValueInMultiValueHeader("Allow", "DELETE", optionsResponse)) {
+            //delete the ldpcv
+            final Response deleteResponse = doDelete(timeMap);
+            assertEquals("Timemap DELETE command failed.", deleteResponse.getStatusCode(), 204);
+            //verify that ldpcv is gone
+            final Response timeMapGet = doGetUnverified(timeMap);
+            assertEquals("Timemap was not deleted", timeMapGet.getStatusCode(), 404);
+            //verify that versioning model is gone
+            final Response getResponse = doGet(versionedResource);
+            confirmAbsenceOfLinkValue(ORIGINAL_RESOURCE_LINK_HEADER, getResponse);
```
awoods: We should probably have an `else` block that adds a "Not Supported" result to the report.
For now, a `TODO: <new GitHub issue>` may be sufficient.

src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpDelete.java
```diff
@@ -53,6 +58,25 @@ public void ldpcvThatAdvertisesDeleteShouldRemoveContainerAndMementos(final Stri
...
+        final String versionedResource = getLocation(response);
+        final String timeMap = getTimeMapUri(response).toString();
+
+        final Response optionsResponse = doOptions(timeMap);
+
+        //check that delete is supported
+        if (hasHeaderValueInMultiValueHeader("Allow", "DELETE", optionsResponse)) {
+            //delete the ldpcv
+            final Response deleteResponse = doDelete(timeMap);
+            assertEquals("Timemap DELETE command failed.", deleteResponse.getStatusCode(), 204);
```
awoods: The `204` assertion already occurs in ``doDelete().

src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpDelete.java
```diff
@@ -53,6 +58,25 @@ public void ldpcvThatAdvertisesDeleteShouldRemoveContainerAndMementos(final Stri
                                         "https://fcrepo.github.io/fcrepo-specification/#ldpcv-delete",
                                         ps);
 
+        //create a versioned resource
+        final Response response = createVersionedResource(uri, info);
+        final String versionedResource = getLocation(response);
+        final String timeMap = getTimeMapUri(response).toString();
```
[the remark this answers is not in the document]
dbernstein: ++

src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpDelete.java
dbernstein: @awoods : Are you sure we want to do this?  It seems to me that perhaps there is a missing "MAY" test (as in "LDPCv may support DELETE" that will fail if delete is not supported.  I believe that is how were are currently doing it with other MAY tests.

src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpDelete.java
dbernstein: ++

PR #146, opened 2018-08-09
Implements tests for org.fcrepo.spec.testsuite.authz.WebACLinking

dbernstein:
Two tests added.  5.4-A is expected to fail since we are not supporting this particular  "MAY" requirement.

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java
@@ -22,9 +22,15 @@ import static org.fcrepo.spec.testsuite.Constants.APPLICATION_SPARQL_UPDATE;
 import static org.fcrepo.spec.testsuite.Constants.BASIC_CONTAINER_BODY;
 import static org.fcrepo.spec.testsuite.Constants.BASIC_CONTAINER_LINK_HEADER;
 import static org.fcrepo.spec.testsuite.Constants.SLUG;
+import static org.testng.AssertJUnit.assertTrue;
 
 import java.io.PrintStream;
 import java.io.StringReader;
+import java.net.URI;
+import java.util.Arrays;
+import java.util.stream.Stream;
+
+import javax.ws.rs.core.Link;
 
 import io.restassured.RestAssured;
 import io.restassured.config.EncoderConfig;
@@ -42,6 +48,7 @@ import org.hamcrest.CoreMatchers;
 import org.hamcrest.Description;
 import org.hamcrest.Matcher;
 import org.hamcrest.Matchers;
+import org.testng.Assert;
 import org.testng.annotations.AfterMethod;
 import org.testng.annotations.BeforeMethod;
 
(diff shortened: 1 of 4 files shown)


whikloj:
No merge access here, but the PR seems fine. Builds and runs as described.

src/main/java/org/fcrepo/spec/testsuite/authz/WebACLinking.java
```diff
@@ -49,9 +51,23 @@ public WebACLinking(final String username, final String password) {
...
+        //create a resource to be used for the acl link
+        final Response aclResponse = createBasicContainer(uri, info.getId() + "-acl");
+        final String aclUri = getLocation(aclResponse);
+
+        //PUT the new Resource
+        final String aclLinkValue = "<" + aclUri + ">; rel=\"acl\"";
+        final Response resource = doPut(uri, Headers.headers(new Header("Link", aclLinkValue)), "test body");
+        final String resourceUri = getLocation(resource);
+        final Response getResource = doGet(resourceUri);
+        confirmPresenceOfLinkValue(aclLinkValue, aclResponse);
```
awoods: This should likely be using `getResource` instead of `aclResponse`.

src/main/java/org/fcrepo/spec/testsuite/authz/WebACLinking.java
```diff
@@ -63,26 +79,29 @@ public void linkToAclOnCreation(final String uri) {
...
+        final Response aclResponse = createBasicContainer(uri, info.getId() + "-acl");
+        final String aclUri = getLocation(aclResponse);
+
+        //PUT the new Resource
+        final String aclLinkValue = "<" + aclUri + ">; rel=\"acl\"";
+        final Response resource = doPutUnverified(uri, Headers.headers(new Header("Link", aclLinkValue)), "test body");
+
+        if (resource.getStatusCode() >= 400) {
+            confirmPresenceOfConstrainedByLink(resource);
+        }
```
awoods: There should probably be an `else` block the verifies response code in the `successRange`.

src/main/resources/testng.xml
```diff
@@ -69,7 +69,9 @@
       <class name="org.fcrepo.spec.testsuite.authz.WebACCrossDomain"/>
       <class name="org.fcrepo.spec.testsuite.authz.WebACGeneral"/>
       <class name="org.fcrepo.spec.testsuite.authz.WebACLinkHeaders"/>
+      -->
```
awoods: Maybe move the `WebACLinking` element above the `Below not yet implemented` section.

src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java
```diff
@@ -354,4 +361,37 @@ public void describeTo(final Description description) {
         return CoreMatchers.both(Matchers.greaterThanOrEqualTo(400)).and(Matchers.lessThan(500));
     }
 
+    protected void confirmPresenceOfLinkValue(final String linkValue, final Response response) {
+        final Link link = Link.valueOf(linkValue);
+        final String relType = link.getRel();
```
awoods: The `relType` variable is never used. I would suggest either removing it, or using it in line:367.

src/main/java/org/fcrepo/spec/testsuite/authz/WebACLinking.java
```diff
@@ -100,6 +105,9 @@ public void conflictIfNotSupportingAclLink(final String uri) {
 
         if (resource.getStatusCode() >= 400) {
             confirmPresenceOfConstrainedByLink(resource);
+        } else {
+            assertTrue(allOf(greaterThanOrEqualTo(200), lessThan(300)).matches(resource.getStatusCode()),
```
awoods: It may be more consistent to use https://github.com/fcrepo/Fedora-API-Test-Suite/blob/master/src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java#L348

...and make that method `protected` instead of `private`.

PR #149, opened 2018-08-16
Adds PUT tests for LDPRv : ie 4.1.2-A,B,C,D

dbernstein:
Test by building and running against a fresh fedora repo.

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java
@@ -207,6 +207,20 @@ public class AbstractTest {
                 .put(uri);
     }
 
+    protected Response doPutUnverified(final String uri, final Headers headers) {
+        return createRequest().headers(headers)
+                              .when()
+                              .put(uri);
+    }
+
+    protected Response doPut(final String uri, final Headers headers) {
+        final Response response = doPutUnverified(uri, headers);
+
+        response.then().statusCode(204);
+
+        return response;
+    }
+
     protected Response doPut(final String uri, final Headers headers, final String body) {
         final Response response = doPutUnverified(uri, headers, body);
 
--- a/src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java
@@ -213,14 +213,6 @@ public class AbstractTest {
                               .put(uri);
     }
 
-    protected Response doPut(final String uri, final Headers headers) {
-        final Response response = doPutUnverified(uri, headers);
-
-        response.then().statusCode(204);
-
-        return response;
-    }
-
     protected Response doPut(final String uri, final Headers headers, final String body) {
         final Response response = doPutUnverified(uri, headers, body);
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java
```diff
@@ -207,6 +207,20 @@ protected Response doPutUnverified(final String uri, final Headers headers, fina
                 .put(uri);
     }
 
+    protected Response doPutUnverified(final String uri, final Headers headers) {
+        return createRequest().headers(headers)
+                              .when()
+                              .put(uri);
+    }
+
+    protected Response doPut(final String uri, final Headers headers) {
```
awoods: This method is never invoked... can be safely removed.

src/main/java/org/fcrepo/spec/testsuite/versioning/AbstractVersioningTest.java
```diff
@@ -80,6 +80,27 @@ protected Response createVersionedResource(final String uri, final TestInfo info
...
+    protected Response putVersionedResourceUnverified(final String uri, final TestInfo info) {
+        return putVersionedResourceUnverified(uri + "/" + info.getId());
+    }
+
+    protected Response putVersionedResourceWithBodyUnverified(final String uri, final TestInfo info,
+                                                              final String body) {
+        return putVersionedResourceWithBodyUnverified(uri + "/" + info.getId(), body);
+    }
+
+    protected Response putVersionedResourceUnverified(final String uri) {
```
awoods: This method can be `private`.

src/main/java/org/fcrepo/spec/testsuite/versioning/LdprvHttpPut.java
```diff
@@ -50,35 +60,57 @@ public void ldprvMustSupportPUT(final String uri) {
...
         final TestInfo info = setupTest("4.1.2-B", "ldprvMustSupportPUTForExistingResources",
                                         "Must support PUT for updating existing LDPRvs",
                                         "https://fcrepo.github.io/fcrepo-specification/#ldprv-put",
                                         ps);
         //create an LDPRv
+        final Response response = putVersionedResourceUnverified(uri, info);
+        assertEquals(response.getStatusCode(), 201);
+
+        final String resourceUri = getLocation(response);
+        final Response getResponse = doGet(resourceUri);
```
awoods: We probably want to specify an `Accept` header of `text/turtle`, since some implementations may return a different RDF serialization by default.

src/main/java/org/fcrepo/spec/testsuite/versioning/LdprvHttpPut.java
```diff
@@ -96,17 +135,22 @@ public void ldpnrvMustSupportPUT(final String uri) {
...
+        assertEquals(response.getStatusCode(), 201);
+        final String resourceUri = getLocation(response);
+        final Response getResponse = doGet(resourceUri);
         //verify that "Allow: PUT" header is present
+        confirmPresenceOfHeaderValueInMultiValueHeader("Allow", "PUT", getResponse);
         //update an existing LDPNRv using a put
-        //verify that it is a TimeGate and has a TimeMap
-
+        final Response response2 = putVersionedResourceWithBodyUnverified(resourceUri,"test2");
+        assertEquals(response2.getStatusCode(), 204);
```
awoods: For completeness, we may as well finish this test with `GET` on the resource to verify that the content has indeed been updated... along the lines of: https://github.com/fcrepo/Fedora-API-Test-Suite/pull/149/files#diff-8817daed51e1756ffb430de8b1f36ae8R105

dbernstein:
@awoods :  updates based on your feedback are now there.

PR #150, opened 2018-08-20
Adds test 4.3.3.1A-D HTTP POST for versioning

bseeger:
This adds versioning tests for POST, tests 4.3.3.1 A-D

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java
@@ -171,6 +171,13 @@ public class AbstractTest {
                 .post(uri);
     }
 
+    protected Response doPostUnverified(final String uri, final String body) {
+        return createRequest()
+                .body(body)
+                .when()
+                .post(uri);
+    }
+
     private Response doPostUnverified(final String uri, final Headers headers) {
         return createRequest().headers(headers)
                 .when()
@@ -190,6 +197,15 @@ public class AbstractTest {
 
         return response;
     }
+
+    protected Response doPost(final String uri, final String body) {
+        final Response response = doPostUnverified(uri, body);
+
+        response.then().statusCode(201);
+        return response;
+    }
+
+
 
     protected Response doPost(final String uri) {
         final Response response = doPostUnverified(uri);
--- a/src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java
@@ -171,7 +171,7 @@ public class AbstractTest {
                 .post(uri);
     }
 
-    protected Response doPostUnverified(final String uri, final String body) {
+    private Response doPostUnverified(final String uri, final String body) {
         return createRequest()
                 .body(body)
                 .when()
--- a/src/main/java/org/fcrepo/spec/testsuite/versioning/AbstractVersioningTest.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/versioning/AbstractVersioningTest.java
@@ -57,7 +57,7 @@ public class AbstractVersioningTest extends AbstractTest {
     protected Response createVersionedResource(final String uri, final TestInfo info) {
         final Headers headers = new Headers(
             new Header("Link", ORIGINAL_RESOURCE_LINK_HEADER),
-            new Header("Content-Type", "text/turtle"));
+            new Header("Content-Type", "text/turtle"));  // this line keeps it from becoming a LDP-NR
         return doPost(uri, headers);
     }
 
--- a/src/main/java/org/fcrepo/spec/testsuite/versioning/AbstractVersioningTest.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/versioning/AbstractVersioningTest.java
@@ -18,6 +18,7 @@
 
 package org.fcrepo.spec.testsuite.versioning;
 
+import static org.fcrepo.spec.testsuite.Constants.CONTENT_DISPOSITION;
 import static org.fcrepo.spec.testsuite.Constants.ORIGINAL_RESOURCE_LINK_HEADER;
 import static org.fcrepo.spec.testsuite.Constants.SLUG;
 
@@ -58,6 +59,14 @@ public class AbstractVersioningTest extends AbstractTest {
             new Header("Link", ORIGINAL_RESOURCE_LINK_HEADER),
             new Header(SLUG, info.getId()));
         return doPost(uri, headers);
+    }
+
+    protected Response createVersionedNonRDFResource(final String uri, final TestInfo info) {
+        final Headers headers = new Headers(
+            new Header("Link", ORIGINAL_RESOURCE_LINK_HEADER),
+            new Header(CONTENT_DISPOSITION, "attachment; filename=\"postNonRDFSource.txt\""),
+            new Header(SLUG, info.getId()));
+        return doPost(uri, headers, "Test String.");
     }
 
     protected Response putVersionedResourceUnverified(final String uri, final TestInfo info) {
@@ -105,6 +114,20 @@ public class AbstractVersioningTest extends AbstractTest {
                                                    return val.equalsIgnoreCase(headerValue);
                                                }).count() > 0;
 
+    }
+
+    protected void confirmPresenceOfVersionLocationHeader(final Response response) {
+        final String location = response.getHeader("Location");
+        Assert.assertTrue(location.contains("fcr:versions"));
+    }
+
+    protected void confirmPresenceOfLinkValue(final String linkValue, final Response response) {
+        final Link link = Link.valueOf(linkValue);
+        final String relType = link.getRel();
+        Assert.assertEquals(getLinksOfRelType(response, link.getRel()).filter(l -> l.equals(link))
+                                                                      .count(),
+                            1,
+                            "Link header with a value of " + linkValue + " must be present but is not!");
     }
 
     protected void confirmAbsenceOfLinkValue(final String linkValue, final Response response) {
--- a/src/main/java/org/fcrepo/spec/testsuite/versioning/AbstractVersioningTest.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/versioning/AbstractVersioningTest.java
@@ -121,14 +121,6 @@ public class AbstractVersioningTest extends AbstractTest {
         Assert.assertTrue(location.contains("fcr:versions"));
     }
 
-    protected void confirmPresenceOfLinkValue(final String linkValue, final Response response) {
-        final Link link = Link.valueOf(linkValue);
-        final String relType = link.getRel();
-        Assert.assertEquals(getLinksOfRelType(response, link.getRel()).filter(l -> l.equals(link))
-                                                                      .count(),
-                            1,
-                            "Link header with a value of " + linkValue + " must be present but is not!");
-    }
 
     protected void confirmAbsenceOfLinkValue(final String linkValue, final Response response) {
         final Link link = Link.valueOf(linkValue);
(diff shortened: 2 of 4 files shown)


src/main/java/org/fcrepo/spec/testsuite/versioning/AbstractVersioningTest.java
```diff
@@ -79,6 +88,20 @@ protected boolean hasHeaderValueInMultiValueHeader(final String headerName, fina
 
     }
 
+    protected void confirmPresenceOfVersionLocationHeader(final Response response) {
+        final String location = response.getHeader("Location");
+        Assert.assertTrue(location.contains("fcr:versions"));
+    }
+
+    protected void confirmPresenceOfLinkValue(final String linkValue, final Response response) {
```
bseeger: I missed that this got moved to AbstractTest.java - I'll remove this function soon.

src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpPost.java
```diff
@@ -54,14 +58,20 @@ public void ldpcvOfLdprsMustSupportPostWithoutMementoDatetimeHeader(final String
                                         "https://fcrepo.github.io/fcrepo-specification/#ldpcv-post",
                                         ps);
 
+        // create the versioned resource
+        final String newResource = createVersionedResource(uri, info).body().asString();
```
awoods: I do not believe we can expect that other Fedora implementations will return the location of the created resource in the body of the response.
You should probably check the `Location` header.

src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpPost.java
```diff
@@ -54,14 +58,20 @@ public void ldpcvOfLdprsMustSupportPostWithoutMementoDatetimeHeader(final String
...
+        // create a memento for the current time
+        final Response versionResponse = doPost(newResource + "/fcr:versions");
```
awoods: I do not believe we can expect that other Fedora implementations will return memento headers in the response to the creation of a new memento.
You should probably make a subsequent `HEAD` or `GET` request in order to check for the memento header.

src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpPost.java
```diff
@@ -73,14 +83,20 @@ public void ldpcvOfLdpnrsMustSupportPostWithoutMementoDatetimeHeader(final Strin
                                         ps);
 
         //same as previous test but with LDP-NR
+        // create the versioned resource
+        final String newResource = createVersionedNonRDFResource(uri, info).body().asString();
```
awoods: see above

src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpPost.java
```diff
@@ -73,14 +83,20 @@ public void ldpcvOfLdpnrsMustSupportPostWithoutMementoDatetimeHeader(final Strin
...
+        // create a memento for the current time
+        final Response versionResponse = doPost(newResource + "/fcr:versions");
+        confirmPresenceOfLinkValue(MEMENTO_LINK_HEADER, versionResponse);
```
awoods: see above

src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpPost.java
```diff
@@ -89,15 +105,20 @@ public void postToldpcvOfLdprsWithoutMementoDatetimeMustIgnoreBody(final String
                                         "Memento-Datetime header MUST ignore any request body.",
                                         "https://fcrepo.github.io/fcrepo-specification/#ldpcv-post",
                                         ps);
-
+        // create the versioned resource
+        final String newResource = createVersionedResource(uri, info).body().asString();
```
awoods: see above

src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpPost.java
```diff
@@ -89,15 +105,20 @@ public void postToldpcvOfLdprsWithoutMementoDatetimeMustIgnoreBody(final String
...
+        // create a memento for the current time
+        final Response versionResponse = doPost(newResource + "/fcr:versions", "Send some random data here");
+        confirmPresenceOfLinkValue(MEMENTO_LINK_HEADER, versionResponse);
```
awoods: see above

src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpPost.java
```diff
@@ -107,6 +128,12 @@ public void postToldpcvOfLdpnrWithoutMementoDatetimeMustIgnoreBody(final String
                                         "https://fcrepo.github.io/fcrepo-specification/#ldpcv-post",
                                         ps);
         //same as previous but with LDP-NR
+        // create the versioned resource
+        final String newResource = createVersionedNonRDFResource(uri, info).body().asString();
```
awoods: see above

src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpPost.java
```diff
@@ -107,6 +128,12 @@ public void postToldpcvOfLdpnrWithoutMementoDatetimeMustIgnoreBody(final String
...
+        // create a memento for the current time
+        final Response versionResponse = doPost(newResource + "/fcr:versions", "Send some random data here");
+        confirmPresenceOfLinkValue(MEMENTO_LINK_HEADER, versionResponse);
```
awoods: see above

src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpPost.java
```diff
@@ -54,14 +58,20 @@ public void ldpcvOfLdprsMustSupportPostWithoutMementoDatetimeHeader(final String
                                         "https://fcrepo.github.io/fcrepo-specification/#ldpcv-post",
                                         ps);
 
+        // create the versioned resource
+        final String newResource = createVersionedResource(uri, info).body().asString();
+        // create a memento for the current time
+        final Response versionResponse = doPost(newResource + "/fcr:versions");
+        confirmPresenceOfLinkValue(MEMENTO_LINK_HEADER, versionResponse);
+        confirmPresenceOfVersionLocationHeader(versionResponse);
```
awoods: Per the requirement being tested, I would expect this test to also be verifying that the state of the created memento is indeed the same the current LDPRv.

src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpPost.java
```diff
@@ -73,14 +83,20 @@ public void ldpcvOfLdpnrsMustSupportPostWithoutMementoDatetimeHeader(final Strin
                                         ps);
 
         //same as previous test but with LDP-NR
+        // create the versioned resource
+        final String newResource = createVersionedNonRDFResource(uri, info).body().asString();
+        // create a memento for the current time
+        final Response versionResponse = doPost(newResource + "/fcr:versions");
+        confirmPresenceOfLinkValue(MEMENTO_LINK_HEADER, versionResponse);
+        confirmPresenceOfVersionLocationHeader(versionResponse);
```
awoods: Per the requirement being tested, I would expect this test to also be verifying that the state of the created memento is indeed the same the current LDPRv.

src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpPost.java
```diff
@@ -89,15 +105,20 @@ public void postToldpcvOfLdprsWithoutMementoDatetimeMustIgnoreBody(final String
                                         "Memento-Datetime header MUST ignore any request body.",
                                         "https://fcrepo.github.io/fcrepo-specification/#ldpcv-post",
                                         ps);
-
+        // create the versioned resource
+        final String newResource = createVersionedResource(uri, info).body().asString();
+        // create a memento for the current time
+        final Response versionResponse = doPost(newResource + "/fcr:versions", "Send some random data here");
+        confirmPresenceOfLinkValue(MEMENTO_LINK_HEADER, versionResponse);
+        confirmPresenceOfVersionLocationHeader(versionResponse);
```
awoods: Per the requirement being tested, I would expect this test to also be verifying that the state of the created memento does not include the `POST`ed body.

src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpPost.java
```diff
@@ -107,6 +128,12 @@ public void postToldpcvOfLdpnrWithoutMementoDatetimeMustIgnoreBody(final String
                                         "https://fcrepo.github.io/fcrepo-specification/#ldpcv-post",
                                         ps);
         //same as previous but with LDP-NR
+        // create the versioned resource
+        final String newResource = createVersionedNonRDFResource(uri, info).body().asString();
+        // create a memento for the current time
+        final Response versionResponse = doPost(newResource + "/fcr:versions", "Send some random data here");
+        confirmPresenceOfLinkValue(MEMENTO_LINK_HEADER, versionResponse);
+        confirmPresenceOfVersionLocationHeader(versionResponse);
```
awoods: Per the requirement being tested, I would expect this test to also be verifying that the state of the created memento does not include the `POST`ed body.

src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java
```diff
@@ -171,6 +171,13 @@ protected Response doPostUnverified(final String uri, final Headers headers, fin
                 .post(uri);
     }
 
+    protected Response doPostUnverified(final String uri, final String body) {
```
awoods: This method can be `private`.

src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpPost.java
dbernstein: @bseeger : there's a getLocation() that should be available.

src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpPost.java
bseeger: Ah, right!  I'm going to wait for Danny's PR to go in and then use his createMemento function here.

bseeger:
note that the LDPCv tests will not past at this time due to this ticket:  https://jira.duraspace.org/browse/FCREPO-2848

src/main/java/org/fcrepo/spec/testsuite/versioning/AbstractVersioningTest.java
```diff
@@ -56,10 +57,18 @@ protected URI getTimeMapUri(final Response response) {
     protected Response createVersionedResource(final String uri, final TestInfo info) {
         final Headers headers = new Headers(
             new Header("Link", ORIGINAL_RESOURCE_LINK_HEADER),
-            new Header(SLUG, info.getId()));
```
bseeger: I don't think we need this unless someone has a good reason to have a known name for a resource.  It makes it hard to run the test multiple times w/o clearing the DB in between.  It's a larger picture question because slugs are used in many spots in the code.

awoods: Agreed. Giving test resources fixed names is not ideal.

PR #151, opened 2018-08-20
Implements tests for 4.2.2-A,B and C

dbernstein:
4.2.2-A,B and C
    4.2.3, 4.2.4, and 4.2.5

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/versioning/AbstractVersioningTest.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/versioning/AbstractVersioningTest.java
@@ -61,10 +61,17 @@ public class AbstractVersioningTest extends AbstractTest {
     }
 
     protected void confirmPresenceOfHeaderValueInMultiValueHeader(final String headerName, final String headerValue,
-                                                                  final Response response) {
+                                                                       final Response response) {
         Assert
             .assertTrue(hasHeaderValueInMultiValueHeader(headerName, headerValue, response),
                         headerName + " with a value of " + headerValue + " must be present but is not!");
+    }
+
+    protected void confirmAbsenceOfHeaderValueInMultiValueHeader(final String headerName, final String headerValue,
+                                                                  final Response response) {
+        Assert
+            .assertFalse(hasHeaderValueInMultiValueHeader(headerName, headerValue, response),
+                        headerName + " with a value of " + headerValue + " must be not present but it is!");
     }
 
     protected boolean hasHeaderValueInMultiValueHeader(final String headerName, final String headerValue,
--- a/src/main/java/org/fcrepo/spec/testsuite/versioning/AbstractVersioningTest.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/versioning/AbstractVersioningTest.java
@@ -61,14 +61,14 @@ public class AbstractVersioningTest extends AbstractTest {
     }
 
     protected void confirmPresenceOfHeaderValueInMultiValueHeader(final String headerName, final String headerValue,
-                                                                       final Response response) {
+                                                                  final Response response) {
         Assert
             .assertTrue(hasHeaderValueInMultiValueHeader(headerName, headerValue, response),
                         headerName + " with a value of " + headerValue + " must be present but is not!");
     }
 
     protected void confirmAbsenceOfHeaderValueInMultiValueHeader(final String headerName, final String headerValue,
-                                                                  final Response response) {
+                                                                 final Response response) {
         Assert
             .assertFalse(hasHeaderValueInMultiValueHeader(headerName, headerValue, response),
                         headerName + " with a value of " + headerValue + " must be not present but it is!");
@@ -93,6 +93,19 @@ public class AbstractVersioningTest extends AbstractTest {
                                                                       .count(),
                             0,
                             "Link header with a value of " + linkValue + " must not be present (but it is)!");
+    }
+
+    /**
+     * Given the URI of the original resource, create a memento based on the
+     * current state of the reosurce.
+     * @param originalResourceUri The resource to be memento-ized.
+     * @return
+     */
+    protected String createMemento(final String originalResourceUri) {
+        final Response response = doGet(originalResourceUri);
+        final URI timeMapURI = getTimeMapUri(response);
+        final Response timeMapResponse = doPost(timeMapURI.toString());
+        return getLocation(timeMapResponse);
     }
 
 }
--- a/src/main/java/org/fcrepo/spec/testsuite/versioning/LdprmHttpOptions.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/versioning/LdprmHttpOptions.java
@@ -17,8 +17,14 @@
  */
 package org.fcrepo.spec.testsuite.versioning;
 
+import static org.testng.Assert.assertEquals;
+
+import java.net.URI;
+
+import io.restassured.response.Response;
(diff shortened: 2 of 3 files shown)


src/main/java/org/fcrepo/spec/testsuite/versioning/LdprmHttpOptions.java
```diff
@@ -88,33 +109,41 @@ public void ldprmOptionsMaySupportDelete(final String uri) {
...
                                         ps);
-
         //create an LDPRm
+        final String mementoURI = createMemento(uri, info);
+        final Response mementoResponse = doGet(mementoURI);
         //verify that a subsequent OPTIONS does not return POST
+        confirmAbsenceOfHeaderValueInMultiValueHeader("Allow", "POST", mementoResponse);
         //verify that issuing a POST returns a 405 (method not allowed).
+        final Response response = doPostUnverified(mementoURI);
+        assertEquals(response.getStatusCode(), 405, "Response to a POST should be 405");
```
bseeger: The 405 seems to be implementation specific  - perhaps it's what's people would use, but I'm not sure if it should be assumed.  Thoughts?

bseeger: It's not specified in the spec, so perhaps this should just check for a 4XX error?

src/main/java/org/fcrepo/spec/testsuite/versioning/LdprmHttpOptions.java
```diff
@@ -123,16 +152,22 @@ public void ldprmMustNotSupportPut(final String uri) {
                                         ps);
 
         //create an LDPRm
+        final String mementoURI = createMemento(uri, info);
+        final Response mementoResponse = doGet(mementoURI);
         //verify that a subsequent OPTIONS does not return PUT
+        confirmAbsenceOfHeaderValueInMultiValueHeader("Allow", "PUT", mementoResponse);
         //verify that issuing a PUT returns a 405 (method not allowed).
+        final Response response = doPutUnverified(mementoURI);
+        assertEquals(response.getStatusCode(), 405, "Response to a PUT should be 405");
```
bseeger: same comment as above about the specific 405 error - should probably just check for a 4xx error.

src/main/java/org/fcrepo/spec/testsuite/versioning/LdprmHttpOptions.java
```diff
@@ -141,8 +176,14 @@ public void ldprmMustNotSupportPatch(final String uri) {
                                         ps);
 
         //create an LDPRm
+        final String mementoURI = createMemento(uri, info);
+        final Response mementoResponse = doGet(mementoURI);
         //verify that a subsequent OPTIONS does not return PATCH
+        confirmAbsenceOfHeaderValueInMultiValueHeader("Allow", "PATCH", mementoResponse);
         //verify that issuing a PATCH returns a 405 (method not allowed).
+        final Response response = doPatchUnverified(mementoURI);
+        assertEquals(response.getStatusCode(), 405, "Response to a PATCH should be 405");
```
bseeger: same comment about the 405 error - should probably just check for a 4xx error.

src/main/java/org/fcrepo/spec/testsuite/versioning/LdprmHttpOptions.java
```diff
@@ -52,16 +58,27 @@ public void ldprmMustSupportOptions(final String uri) {
                                         ps);
 
         //create an LDPRm
+        final String mementoURI = createMemento(uri, info);
+        final Response mementoResponse = doGet(mementoURI);
         //verify that a subsequent GET returns OPTIONS in "Allow" header.
+        confirmPresenceOfHeaderValueInMultiValueHeader("Allow", "OPTIONS", mementoResponse);
+    }
 
+    private String createMemento(final String uri, final TestInfo info) {
```
bseeger: I think something like this should go into the AbstractVersioningTest.java file.  But, I don't think it should do things - namely create a versioned resource and create a memento.  I'd say pass in the URI of the versioned resource, and only have this create the memento.  

I'd like to use this function in my PR, hence wanting it up in the Abstract Versioning class.  It's pretty handy.

src/main/java/org/fcrepo/spec/testsuite/versioning/AbstractVersioningTest.java
```diff
@@ -61,12 +61,19 @@ protected Response createVersionedResource(final String uri, final TestInfo info
     }
 
     protected void confirmPresenceOfHeaderValueInMultiValueHeader(final String headerName, final String headerValue,
-                                                                  final Response response) {
+                                                                       final Response response) {
```
bseeger: should revert this change.

src/main/java/org/fcrepo/spec/testsuite/versioning/AbstractVersioningTest.java
```diff
@@ -61,12 +61,19 @@ protected Response createVersionedResource(final String uri, final TestInfo info
...
         Assert
             .assertTrue(hasHeaderValueInMultiValueHeader(headerName, headerValue, response),
                         headerName + " with a value of " + headerValue + " must be present but is not!");
     }
 
+    protected void confirmAbsenceOfHeaderValueInMultiValueHeader(final String headerName, final String headerValue,
+                                                                  final Response response) {
```
bseeger: picky, I know, but there's an extra space on this line before 'final'  :)

src/main/java/org/fcrepo/spec/testsuite/versioning/AbstractVersioningTest.java
dbernstein: @bseeger :  good catch.

src/main/java/org/fcrepo/spec/testsuite/versioning/AbstractVersioningTest.java
dbernstein: picky is good.

dbernstein: picky is good.

src/main/java/org/fcrepo/spec/testsuite/versioning/LdprmHttpOptions.java
dbernstein: @bseeger 👍

src/main/java/org/fcrepo/spec/testsuite/versioning/LdprmHttpOptions.java
dbernstein: @bseeger : I'm not sure either.   It seems to me that anything other than 405 would indicate some other problem.  I would suggest that we keep it as 405 and change it if we run into complaints from one or more Fedora implementers.

dbernstein:
@bseeger : feel free to merged it when you're ready.

src/main/java/org/fcrepo/spec/testsuite/versioning/LdprmHttpOptions.java
awoods: I agree that 405 is the appropriate response here. 
As a point of reference for other tests that need to support 4xx status codes, there is this helper:
https://github.com/fcrepo/Fedora-API-Test-Suite/blob/master/src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java#L360

Issue #155, opened 2018-09-10
Implement tests in ExternalBinaryContent (section 3.9)


bbpennel:
Implement external binary content tests in for section 3.9

PR #165, opened 2018-09-12
155 - Adds tests for External Binary Content

bbpennel:
Addresses #155 

* Adds external binary tests
* Some changes to scope of unverified method calls to allow for direct usage in tests.
* Adds mock web server for use in tests.

Fedora must be run with an allow list that permits access to your temp directory and http uris. I tested using:
`mvn jetty:run -Dfcrepo.external.content.allowed=/path/to/fcrepo4/fcrepo-http-api/src/test/resources/allowed_external_paths.txt`

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
@@ -22,6 +22,7 @@ import static org.fcrepo.spec.testsuite.Constants.DIGEST;
 import static org.fcrepo.spec.testsuite.Constants.EXTERNAL_CONTENT_LINK_REL;
 import static org.fcrepo.spec.testsuite.Constants.LINK;
 import static org.fcrepo.spec.testsuite.Constants.SLUG;
+import static org.fcrepo.spec.testsuite.Constants.NON_RDF_SOURCE_INTERACTION_MODEL;
 import java.io.File;
 import java.io.IOException;
 import java.util.Arrays;
@@ -53,7 +54,7 @@ import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.anyOf;
 import static org.hamcrest.Matchers.greaterThanOrEqualTo;
 import static org.testng.Assert.assertTrue;
-
+import static org.testng.AssertJUnit.assertTrue;
 import static java.util.Collections.emptyList;
 
 /**
@@ -87,7 +88,7 @@ public class ExternalBinaryContent extends AbstractTest {
         wireMockServer = new WireMockServer(options().dynamicPort());
         wireMockServer.start();
         externalUri = mockHttpResource("file.txt", "text/plain", "binary content");
-        externalUriWithNoType = mockHttpResourceNoType("notype2", "other content");
+        externalUriWithNoType = mockHttpResourceNoType("notype", "other content");
 
         final File tempFile = File.createTempFile("ext", null);
         tempFile.deleteOnExit();
@@ -124,7 +125,7 @@ public class ExternalBinaryContent extends AbstractTest {
     }
 
     /**
-     * 3.9-A-1b PostCreate Copy
+     * 3.9-A-1b PostCreate Copy File URI
      *
      * @param uri of base container of Fedora server
      */
@@ -165,7 +166,7 @@ public class ExternalBinaryContent extends AbstractTest {
     }
 
     /**
-     * 3.9-A-2b PostCreate Redirect
+     * 3.9-A-2b PostCreate Redirect File URI
      *
      * @param uri of base container of Fedora server
      */
@@ -205,7 +206,7 @@ public class ExternalBinaryContent extends AbstractTest {
     }
 
     /**
-     * 3.9-A-3b PostCreate Proxy
+     * 3.9-A-3b PostCreate Proxy File URI
      *
      * @param uri of base container of Fedora server
      */
@@ -246,7 +247,7 @@ public class ExternalBinaryContent extends AbstractTest {
     }
 
     /**
-     * 3.9-B-1b PutCreate Copy
+     * 3.9-B-1b PutCreate Copy File URI
      *
      * @param uri of base container of Fedora server
      */
@@ -288,7 +289,7 @@ public class ExternalBinaryContent extends AbstractTest {
     }
 
     /**
-     * 3.9-B-2b PutCreate Redirect
+     * 3.9-B-2b PutCreate Redirect File URI
      *
      * @param uri of base container of Fedora server
      */
@@ -330,7 +331,7 @@ public class ExternalBinaryContent extends AbstractTest {
     }
 
     /**
-     * 3.9-B-3b PutCreate Proxy
+     * 3.9-B-3b PutCreate Proxy File URI
      *
      * @param uri of base container of Fedora server
      */
@@ -364,11 +365,7 @@ public class ExternalBinaryContent extends AbstractTest {
                         "rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
 
-        final Headers headers = new Headers(
-                new Header(CONTENT_DISPOSITION, "attachment; filename=\"postCreate.txt\""),
-                new Header(SLUG, info.getId()));
-        final Response resource = doPost(uri, headers, "TestString.");
-        final String locationHeader = getLocation(resource);
+        final String locationHeader = createNonRdfSource(uri, info.getId());
 
         final Headers headers2 = new Headers(
                 new Header(LINK, externalContentHeader(externalUri, "copy")));
@@ -376,7 +373,7 @@ public class ExternalBinaryContent extends AbstractTest {
     }
 
     /**
-     * 3.9-C-1b PutUpdate Copy
+     * 3.9-C-1b PutUpdate Copy File URI
      *
      * @param uri of base container of Fedora server
      */
@@ -389,11 +386,7 @@ public class ExternalBinaryContent extends AbstractTest {
                         "rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
 
-        final Headers headers = new Headers(
-                new Header(CONTENT_DISPOSITION, "attachment; filename=\"postCreate.txt\""),
-                new Header(SLUG, info.getId()));
-        final Response resource = doPost(uri, headers, "TestString.");
-        final String locationHeader = getLocation(resource);
+        final String locationHeader = createNonRdfSource(uri, info.getId());
 
         final Headers headers2 = new Headers(
                 new Header(LINK, externalContentHeader(externalFileUri, "copy")));
@@ -414,11 +407,7 @@ public class ExternalBinaryContent extends AbstractTest {
                         "rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
 
-        final Headers headers = new Headers(
-                new Header(CONTENT_DISPOSITION, "attachment; filename=\"postCreate.txt\""),
-                new Header(SLUG, info.getId()));
-        final Response resource = doPost(uri, headers, "TestString.");
-        final String locationHeader = getLocation(resource);
+        final String locationHeader = createNonRdfSource(uri, info.getId());
 
         final Headers headers2 = new Headers(
                 new Header(LINK, externalContentHeader(externalUri, "redirect")));
@@ -426,7 +415,7 @@ public class ExternalBinaryContent extends AbstractTest {
     }
 
     /**
-     * 3.9-C-2b PutUpdate Redirect
+     * 3.9-C-2b PutUpdate Redirect File URI
      *
      * @param uri of base container of Fedora server
      */
@@ -439,11 +428,7 @@ public class ExternalBinaryContent extends AbstractTest {
                         "rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
 
-        final Headers headers = new Headers(
-                new Header(CONTENT_DISPOSITION, "attachment; filename=\"postCreate.txt\""),
-                new Header(SLUG, info.getId()));
-        final Response resource = doPost(uri, headers, "TestString.");
-        final String locationHeader = getLocation(resource);
+        final String locationHeader = createNonRdfSource(uri, info.getId());
 
         final Headers headers2 = new Headers(
                 new Header(LINK, externalContentHeader(externalFileUri, "redirect")));
@@ -464,11 +449,7 @@ public class ExternalBinaryContent extends AbstractTest {
                         "rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
 
-        final Headers headers = new Headers(
-                new Header(CONTENT_DISPOSITION, "attachment; filename=\"postCreate.txt\""),
-                new Header(SLUG, info.getId()));
-        final Response resource = doPost(uri, headers, "TestString.");
-        final String locationHeader = getLocation(resource);
+        final String locationHeader = createNonRdfSource(uri, info.getId());
 
         final Headers headers2 = new Headers(
                 new Header(LINK, externalContentHeader(externalUri, "proxy")));
@@ -476,7 +457,7 @@ public class ExternalBinaryContent extends AbstractTest {
     }
 
     /**
-     * 3.9-C-3b PutUpdate Proxy
+     * 3.9-C-3b PutUpdate Proxy File URI
      *
      * @param uri of base container of Fedora server
      */
@@ -489,11 +470,7 @@ public class ExternalBinaryContent extends AbstractTest {
                         "rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
 
-        final Headers headers = new Headers(
-                new Header(CONTENT_DISPOSITION, "attachment; filename=\"postCreate.txt\""),
-                new Header(SLUG, info.getId()));
-        final Response resource = doPost(uri, headers, "TestString.");
-        final String locationHeader = getLocation(resource);
+        final String locationHeader = createNonRdfSource(uri, info.getId());
 
         final Headers headers2 = new Headers(
                 new Header(LINK, externalContentHeader(externalFileUri, "proxy")));
@@ -742,6 +719,7 @@ public class ExternalBinaryContent extends AbstractTest {
                 new Header(SLUG, info.getId()));
         final String location = getLocation(doPost(uri, headers));
 
+
         // Type attribute supersedes Content-Type from external server
         doHead(location).then()
                 .header("Content-Type", "text/special");
@@ -784,6 +762,12 @@ public class ExternalBinaryContent extends AbstractTest {
                         "the request entity must reject request if it cannot guarantee all of the response headers " +
                         "required by the LDP-NR interaction model in this specification.",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final String location = createExternalBinary(uri, info.getId(), "proxy", null);
+
+        final Response resp = doGet(location);
+        assertTrue("Response does not contain link of rel type = describedby",
+                getLinksOfRelType(resp, "describedby").count() > 0);
     }
 
     /**
@@ -799,6 +783,10 @@ public class ExternalBinaryContent extends AbstractTest {
                         "the request entity must reject request if it cannot guarantee all of the response headers " +
                         "required by the LDP-NR interaction model in this specification.",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final String location = createExternalBinary(uri, info.getId(), "proxy", "text/plain");
+
+        doGet(location).then().header("Content-Type", "text/plain");
     }
 
     /**
@@ -814,6 +802,10 @@ public class ExternalBinaryContent extends AbstractTest {
                         "the request entity must reject request if it cannot guarantee all of the response headers " +
                         "required by the LDP-NR interaction model in this specification.",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final String location = createExternalBinary(uri, info.getId(), "proxy", "text/plain");
+
+        doGet(location).then().header("Content-Length", is(notNullValue()));
     }
 
     /**
@@ -829,6 +821,13 @@ public class ExternalBinaryContent extends AbstractTest {
                         "the request entity must reject request if it cannot guarantee all of the response headers " +
                         "required by the LDP-NR interaction model in this specification.",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final String location = createExternalBinary(uri, info.getId(), "proxy", "text/plain");
+
+        final Response resp = doGet(location);
+        assertTrue(getLinksOfRelTypeAsUris(resp, "type")
+                .anyMatch(p -> p.toString().equals(NON_RDF_SOURCE_INTERACTION_MODEL)),
+                "Interaction model link header of rel=type is required");
     }
 
     /**
@@ -861,10 +860,7 @@ public class ExternalBinaryContent extends AbstractTest {
                         "\"Want-Digest\" header.",
                 "https://fedora.info/2018/06/25/spec/#redirect-and-proxy", ps);
 
-        final Headers headers = new Headers(
-                new Header(LINK, externalContentHeader(externalUri, "redirect")),
-                new Header(SLUG, info.getId()));
-        final String location = getLocation(doPost(uri, headers));
+        final String location = createExternalBinary(uri, info.getId(), "redirect", null);
 
         final String checksum = "md5;q=0.3,sha;q=1";
         final Response wantDigestResponse = doGetUnverified(location, new Header("Want-Digest", checksum));
@@ -887,10 +883,7 @@ public class ExternalBinaryContent extends AbstractTest {
                         "\"Want-Digest\" header.",
                 "https://fedora.info/2018/06/25/spec/#redirect-and-proxy", ps);
 
-        final Headers headers = new Headers(
-                new Header(LINK, externalContentHeader(externalUri, "proxy")),
-                new Header(SLUG, info.getId()));
-        final String location = getLocation(doPost(uri, headers));
+        final String location = createExternalBinary(uri, info.getId(), "proxy", null);
 
         final String checksum = "md5;q=0.3,sha;q=1";
         final Response wantDigestResponse = doGet(location, new Header("Want-Digest", checksum));
@@ -940,7 +933,7 @@ public class ExternalBinaryContent extends AbstractTest {
                 new Header(SLUG, info.getId()));
         final String location = getLocation(doPostUnverified(uri, headers));
 
-        doHeadUnverified(location).then()
+        doHeadUnverified(location, true).then()
                 .statusCode(anyOf(is(302), is(307)));
     }
 
@@ -1000,5 +993,21 @@ public class ExternalBinaryContent extends AbstractTest {
             link += "; type=\"" + type + "\"";
         }
         return link;
+    }
+
+    private String createExternalBinary(final String uri, final String id, final String handling,
+            final String contentType) {
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalUri, handling, contentType)),
+                new Header(SLUG, id));
+        return getLocation(doPost(uri, headers));
+    }
+
+    private String createNonRdfSource(final String uri, final String id) {
+        final Headers headers = new Headers(
+                new Header(CONTENT_DISPOSITION, "attachment; filename=\"postCreate.txt\""),
+                new Header(SLUG, id));
+        final Response resource = doPost(uri, headers, "TestString.");
+        return getLocation(resource);
     }
 }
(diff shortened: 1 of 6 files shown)


src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -112,6 +280,33 @@ public void putCreateExternalBinaryContentRedirect(final String uri) {
...
+    @Test(groups = { "SHOULD" })
+    @Parameters({ "param1" })
+    public void putCreateExternalFileUriBinaryRedirect(final String uri) {
+        final TestInfo info = setupTest("3.9-B-2b", "putCreateExternalFileUriBinaryRedirect",
+                "Fedora servers should support the creation of LDP-NRs with content external to the " +
+                        "request entity, as indicated by a link with " +
+                        "rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"",
+                "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final String location = uri + info.getId();
```
dbernstein: You need a slash after uri: Should be `uri + "/" + info.getId();`

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -319,21 +713,40 @@ public void binaryContentContentTypeAndType(final String uri) {
                         "type of the external content, if provided. Any Content-Type header in the request should be " +
                         "ignored.",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final Headers headers = new Headers(
```
awoods: I believe this element of the specification is referring to `Content-Type` in the creation request. Therefore, I would suggest adding a `Content-Type` header to the `headers` list.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -127,6 +322,33 @@ public void putCreateExternalBinaryContentProxy(final String uri) {
                         "request entity, as indicated by a link with " +
                         "rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final String location = uri + info.getId();
```
dbernstein: It would be better if you check to see if the base URL does not end with a `/`. If so add in a slash.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -127,6 +322,33 @@ public void putCreateExternalBinaryContentProxy(final String uri) {
...
+    @Test(groups = { "SHOULD" })
+    @Parameters({ "param1" })
+    public void putCreateExternalFileUriBinaryProxy(final String uri) {
+        final TestInfo info = setupTest("3.9-B-3b", "putCreateExternalFileUriBinaryProxy",
+                "Fedora servers should support the creation of LDP-NRs with content external to the " +
+                        "request entity, as indicated by a link with " +
+                        "rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"",
+                "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final String location = uri + info.getId();
```
dbernstein: It would be better if you check to see if the base URL does not end with a `/`. If so add in a slash.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -319,21 +713,40 @@ public void binaryContentContentTypeAndType(final String uri) {
...
     public void binaryContentContentTypeAndNoType(final String uri) {
         final TestInfo info = setupTest("3.9-F-6", "binaryContentContentTypeAndNoType",
                 "Fedora servers must use the value of the type attribute in the external content link as the media " +
                         "type of the external content, if provided. Any Content-Type header in the request should be " +
                         "ignored.",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        // Providing external content which does not return Content-Type
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalUriWithNoType, "proxy", "text/special")),
```
awoods: It is good to verify the behavior in this test and the previous, but I believe the specification is referring to `Content-Type` in the initial creation request.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -349,6 +762,12 @@ public void binaryContentGuaranteeHeadersDescribedBy(final String uri) {
                         "the request entity must reject request if it cannot guarantee all of the response headers " +
                         "required by the LDP-NR interaction model in this specification.",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final String location = createExternalBinary(uri, info.getId(), "proxy", null);
+
+        final Response resp = doGet(location);
+        assertTrue("Response does not contain link of rel type = describedby",
```
awoods: We likely want to add more expected headers, but that can be as a follow-on ticket.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -364,6 +783,10 @@ public void binaryContentGuaranteeHeadersContentType(final String uri) {
                         "the request entity must reject request if it cannot guarantee all of the response headers " +
                         "required by the LDP-NR interaction model in this specification.",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final String location = createExternalBinary(uri, info.getId(), "proxy", "text/plain");
+
+        doGet(location).then().header("Content-Type", "text/plain");
```
awoods: Can probably merge this test with the previous.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -379,6 +802,10 @@ public void binaryContentGuaranteeHeadersContentLength(final String uri) {
                         "the request entity must reject request if it cannot guarantee all of the response headers " +
                         "required by the LDP-NR interaction model in this specification.",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final String location = createExternalBinary(uri, info.getId(), "proxy", "text/plain");
+
+        doGet(location).then().header("Content-Length", is(notNullValue()));
```
awoods: Can probably merge this test with the previous.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -394,6 +821,13 @@ public void binaryContentGuaranteeHeadersInteractionModel(final String uri) {
                         "the request entity must reject request if it cannot guarantee all of the response headers " +
                         "required by the LDP-NR interaction model in this specification.",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final String location = createExternalBinary(uri, info.getId(), "proxy", "text/plain");
+
+        final Response resp = doGet(location);
+        assertTrue(getLinksOfRelTypeAsUris(resp, "type")
```
awoods: Can probably merge this test with the previous.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -52,6 +116,33 @@ public void postCreateExternalBinaryContentCopy(final String uri) {
                         "request entity, as indicated by a link with " +
                         "rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalUri, "copy")),
+                new Header(SLUG, info.getId()));
```
dbernstein: Please avoid using the `Slug` header where possible as support for it is not required AFAIK.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -52,6 +116,33 @@ public void postCreateExternalBinaryContentCopy(final String uri) {
...
+    public void postCreateExternalFileUriBinaryCopy(final String uri) {
+        final TestInfo info = setupTest("3.9-A-1b", "postCreateExternalFileUriBinaryCopy",
+                "Fedora servers should support the creation of LDP-NRs with content external to the " +
+                        "request entity, as indicated by a link with " +
+                        "rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"",
+                "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalFileUri, "copy")),
+                new Header(SLUG, info.getId()));
```
dbernstein: Please avoid using the `Slug` header where possible as support for it is not required AFAIK.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -67,6 +158,31 @@ public void postCreateExternalBinaryContentRedirect(final String uri) {
                         "request entity, as indicated by a link with " +
                         "rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalUri, "redirect")),
+                new Header(SLUG, info.getId()));
```
dbernstein: Please avoid using the `Slug` header where possible as support for it is not required AFAIK.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -67,6 +158,31 @@ public void postCreateExternalBinaryContentRedirect(final String uri) {
...
+    public void postCreateExternalFileUriBinaryRedirect(final String uri) {
+        final TestInfo info = setupTest("3.9-A-2b", "postCreateExternalFileUriBinaryRedirect",
+                "Fedora servers should support the creation of LDP-NRs with content external to the " +
+                        "request entity, as indicated by a link with " +
+                        "rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"",
+                "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalFileUri, "redirect")),
+                new Header(SLUG, info.getId()));
```
dbernstein: Please avoid using the `Slug` header where possible as support for it is not required AFAIK.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -82,6 +198,31 @@ public void postCreateExternalBinaryContentProxy(final String uri) {
                         "request entity, as indicated by a link with " +
                         "rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalUri, "proxy")),
+                new Header(SLUG, info.getId()));
```
dbernstein: Please avoid using the `Slug` header where possible as support for it is not required AFAIK.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -422,6 +859,15 @@ public void binaryContentRedirectWantDigest(final String uri) {
...
                 "https://fedora.info/2018/06/25/spec/#redirect-and-proxy", ps);
+
+        final String location = createExternalBinary(uri, info.getId(), "redirect", null);
+
+        final String checksum = "md5;q=0.3,sha;q=1";
+        final Response wantDigestResponse = doGetUnverified(location, new Header("Want-Digest", checksum));
+
+        final Headers responseHeaders = wantDigestResponse.getHeaders();
+        assertTrue(responseHeaders.getValue(DIGEST).contains("md5") ||
+                responseHeaders.getValue(DIGEST).contains("sha"), "OK");
```
awoods: The `"OK"` message may be misleading, as the message is only displayed on failure. I would suggest removing the message, or adding a message that says something like, "Expected Want-Digest value not found".

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -82,6 +198,31 @@ public void postCreateExternalBinaryContentProxy(final String uri) {
...
+    public void postCreateExternalFileUriBinaryProxy(final String uri) {
+        final TestInfo info = setupTest("3.9-A-3b", "postCreateExternalFileUriBinaryProxy",
+                "Fedora servers should support the creation of LDP-NRs with content external to the " +
+                        "request entity, as indicated by a link with " +
+                        "rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"",
+                "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalFileUri, "proxy")),
+                new Header(SLUG, info.getId()));
```
dbernstein: Please avoid using the `Slug` header where possible as support for it is not required AFAIK.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -186,6 +489,17 @@ public void unsupportedExternalBinaryContentStatus(final String uri) {
...
                         "with a 4xx range status code",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        if (externalContentSupported(uri)) {
+            return;
+        }
+
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalUri, "proxy")),
+                new Header(SLUG, info.getId()));
```
dbernstein: Please avoid using the `Slug` header where possible as support for it is not required AFAIK.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -201,6 +515,16 @@ public void unsupportedExternalBinaryContentConstraint(final String uri) {
...
                         "rel=\"http://www.w3.org/ns/ldp#constrainedBy\" link in the Link response header.",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        if (externalContentSupported(uri)) {
+            return;
+        }
+
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalUri, "proxy")),
+                new Header(SLUG, info.getId()));
```
dbernstein: Please avoid using the `Slug` header where possible as support for it is not required AFAIK.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -422,6 +859,15 @@ public void binaryContentRedirectWantDigest(final String uri) {
...
                         "\"Want-Digest\" header.",
                 "https://fedora.info/2018/06/25/spec/#redirect-and-proxy", ps);
+
+        final String location = createExternalBinary(uri, info.getId(), "redirect", null);
+
+        final String checksum = "md5;q=0.3,sha;q=1";
+        final Response wantDigestResponse = doGetUnverified(location, new Header("Want-Digest", checksum));
+
+        final Headers responseHeaders = wantDigestResponse.getHeaders();
+        assertTrue(responseHeaders.getValue(DIGEST).contains("md5") ||
```
awoods: We may want to expand the list of accepted digest algorithms to include: sha256 and sha512.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -436,6 +882,15 @@ public void binaryContentProxyWantDigest(final String uri) {
                 "Fedora servers supporting \"redirect\" external content types MUST correctly respond to the " +
                         "\"Want-Digest\" header.",
                 "https://fedora.info/2018/06/25/spec/#redirect-and-proxy", ps);
+
+        final String location = createExternalBinary(uri, info.getId(), "proxy", null);
+
+        final String checksum = "md5;q=0.3,sha;q=1";
```
awoods: We may want to expand the list of accepted digest algorithms to include: sha256 and sha512.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -464,6 +927,87 @@ public void binaryContentRedirectStatusHead(final String uri) {
...
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalUri, handling, contentType)),
+                new Header(SLUG, id));
+        return getLocation(doPost(uri, headers));
+    }
+
+    private String createNonRdfSource(final String uri, final String id) {
+        final Headers headers = new Headers(
+                new Header(CONTENT_DISPOSITION, "attachment; filename=\"postCreate.txt\""),
+                new Header(SLUG, id));
```
dbernstein: Please avoid using the `Slug` header where possible as support for it is not required AFAIK.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -464,6 +927,87 @@ public void binaryContentRedirectStatusHead(final String uri) {
...
+            link += "; type=\"" + type + "\"";
+        }
+        return link;
+    }
+
+    private String createExternalBinary(final String uri, final String id, final String handling,
+            final String contentType) {
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalUri, handling, contentType)),
+                new Header(SLUG, id));
```
dbernstein: Please avoid using the `Slug` header where possible as support for it is not required AFAIK.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -464,6 +927,87 @@ public void binaryContentRedirectStatusHead(final String uri) {
                 "A successful response to a HEAD request for external content with handling of redirect " +
                         "must have status code of either 302 (Found) or 307 (Temporary Redirect)",
                 "https://fedora.info/2018/06/25/spec/#redirect-and-proxy", ps);
+
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalUri, "redirect")),
+                new Header(SLUG, info.getId()));
```
dbernstein: Please avoid using the `Slug` header where possible as support for it is not required AFAIK.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -450,6 +905,14 @@ public void binaryContentRedirectStatusGet(final String uri) {
                 "A successful response to a GET request for external content with handling of redirect " +
                         "must have status code of either 302 (Found) or 307 (Temporary Redirect)",
                 "https://fedora.info/2018/06/25/spec/#redirect-and-proxy", ps);
+
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalUri, "redirect")),
+                new Header(SLUG, info.getId()));
```
dbernstein: Please avoid using the `Slug` header where possible as support for it is not required AFAIK.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -319,21 +713,40 @@ public void binaryContentContentTypeAndType(final String uri) {
...
         final TestInfo info = setupTest("3.9-F-6", "binaryContentContentTypeAndNoType",
                 "Fedora servers must use the value of the type attribute in the external content link as the media " +
                         "type of the external content, if provided. Any Content-Type header in the request should be " +
                         "ignored.",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        // Providing external content which does not return Content-Type
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalUriWithNoType, "proxy", "text/special")),
+                new Header(SLUG, info.getId()));
```
dbernstein: Please avoid using the `Slug` header where possible as support for it is not required AFAIK.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -319,21 +713,40 @@ public void binaryContentContentTypeAndType(final String uri) {
                         "type of the external content, if provided. Any Content-Type header in the request should be " +
                         "ignored.",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalUri, "proxy", "text/special")),
+                new Header(SLUG, info.getId()));
```
dbernstein: Please avoid using the `Slug` header where possible as support for it is not required AFAIK.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -301,9 +689,15 @@ public void binaryContentNoTypeDefault(final String uri) {
...
                 "Fedora servers must use the value of the type attribute in the external content link as the media " +
-                        "type of the external content, if provided. Servers may reject the request with a 4xx range " +
-                        "status code.",
+                        "type of the external content, if provided. If there is no type attribute: " +
+                        "Servers may reject the request with a 4xx range status code.",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalUriWithNoType, "proxy")),
+                new Header(SLUG, info.getId()));
```
dbernstein: Please avoid using the `Slug` header where possible as support for it is not required AFAIK.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -287,8 +665,18 @@ public void binaryContentNoTypeExternalType(final String uri) {
...
         final TestInfo info = setupTest("3.9-F-3", "binaryContentNoTypeDefault",
                 "Fedora servers must use the value of the type attribute in the external content link as the media " +
-                        "type of the external content, if provided. Servers may use a default media type.",
+                        "type of the external content, if provided. If there is no type attribute: " +
+                        "Servers may use a default media type.",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalUriWithNoType, "proxy")),
+                new Header(SLUG, info.getId()));
```
dbernstein: Please avoid using the `Slug` header where possible as support for it is not required AFAIK.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -271,10 +641,18 @@ public void binaryContentMediaType(final String uri) {
...
-                        "accessing the external content via the specified scheme (e.g. the Content-Type header for " +
-                        "external content accessed via http).",
+                        "type of the external content, if provided. If there is no type attribute: " +
+                        "Servers may use the media type obtained when accessing the external content via the " +
+                        "specified scheme (e.g. the Content-Type header for external content accessed via http).",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalUri, "proxy")),
+                new Header(SLUG, info.getId()));
```
dbernstein: Please avoid using the `Slug` header where possible as support for it is not required AFAIK.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -259,6 +620,15 @@ public void binaryContentMediaType(final String uri) {
                 "Fedora servers must use the value of the type attribute in the external content link as the media " +
                         "type of the external content, if provided.",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalUriWithNoType, "proxy", "text/special")),
+                new Header(SLUG, info.getId()));
```
dbernstein: Please avoid using the `Slug` header where possible as support for it is not required AFAIK.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -245,6 +598,14 @@ public void unsupportedBinaryContentHandlingAttribute(final String uri) {
                         "to fail must be described in a resource indicated by a " +
                         "rel=\"http://www.w3.org/ns/ldp#constrainedBy\" link in the Link response header.",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalUri, "unsupported")),
+                new Header(SLUG, info.getId()));
```
dbernstein: Please avoid using the `Slug` header where possible as support for it is not required AFAIK.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -230,6 +575,14 @@ public void externalBinaryContentHandlingStatus(final String uri) {
                 "Fedora servers must reject with a 4xx range status code requests for which the handling attribute " +
                         "is not present or cannot be respected.",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final Headers headers = new Headers(
+                new Header(LINK, externalContentHeader(externalUri, null)),
+                new Header(SLUG, info.getId()));
```
dbernstein: Please avoid using the `Slug` header where possible as support for it is not required AFAIK.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -216,6 +540,27 @@ public void externalBinaryContentHandlingAttribute(final String uri) {
...
+                .collect(Collectors.toList());
+
+        assertThat("Accept-External-Content-Handling header contains at least one expected handling",
+                acceptedAndExpected.size(), greaterThanOrEqualTo(1));
+
+        // Test that at least one of the handlings specified by the serve is supported
+        for (final String handling : acceptedAndExpected) {
+            final Headers headers = new Headers(
+                    new Header(LINK, externalContentHeader(externalUri, handling)),
+                    new Header(SLUG, info.getId()));
```
dbernstein: Please avoid using the `Slug` header where possible as support for it is not required AFAIK.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -216,6 +540,27 @@ public void externalBinaryContentHandlingAttribute(final String uri) {
...
+        final List<String> acceptedHandlingTypes = getAcceptedHandlingTypes(uri);
+
+        final List<String> acceptedAndExpected = acceptedHandlingTypes.stream()
+                .filter(HANDLINGS::contains)
+                .collect(Collectors.toList());
+
+        assertThat("Accept-External-Content-Handling header contains at least one expected handling",
+                acceptedAndExpected.size(), greaterThanOrEqualTo(1));
+
+        // Test that at least one of the handlings specified by the serve is supported
```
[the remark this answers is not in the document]
bbpennel: hmm, i feel like that might be a little confusing looking with the "At least one of the following handling attributes must be supported" text right above it, but I could change it to verify all the handlings returned by the server work.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
bbpennel: Sure, that seems reasonable, there is a lot of redundancy at the moment

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
bbpennel: This was mainly borrowed from https://github.com/fcrepo/Fedora-API-Test-Suite/blob/master/src/main/java/org/fcrepo/spec/testsuite/crud/HttpGet.java#L228-L254 but i'm realizing now that that was a more complicated case than I need here,  I can probably mimic https://github.com/fcrepo/Fedora-API-Test-Suite/blob/master/src/main/java/org/fcrepo/spec/testsuite/crud/HttpGet.java#L202-L218 instead

bbpennel: And I just also realized that you were referring to the message displayed when the assertion fails :) I keep forgetting that with testng the message goes after the condition. Too much time in junit land.

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
bbpennel: Okay, I definitely misinterpreted that, will fix

src/main/java/org/fcrepo/spec/testsuite/crud/ExternalBinaryContent.java
```diff
@@ -97,6 +238,33 @@ public void putCreateExternalBinaryContentCopy(final String uri) {
                         "request entity, as indicated by a link with " +
                         "rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"",
                 "https://fcrepo.github.io/fcrepo-specification/#external-content", ps);
+
+        final String location = joinLocation(uri, info.getId());
```
awoods: hmm... since we are talking about the URI passed in from `param1`, which is used in all of the tests, I would suggest that ensuring the `uri` ends in `/` be handled once, here: https://github.com/fcrepo/Fedora-API-Test-Suite/blob/master/src/main/java/org/fcrepo/spec/testsuite/TestSuiteGlobals.java#L58
..which is invoked here: https://github.com/fcrepo/Fedora-API-Test-Suite/blob/master/src/main/java/org/fcrepo/spec/testsuite/App.java#L101

PR #170, opened 2018-09-13
152 - Cleanup resources created by tests

bbpennel:
Deletes all resources created by tests
* behavior attempts to account for possibility that a test container was not created, and whether or not the implementation supports recursive deletion of containers with children.

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/TestSuiteGlobals.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/TestSuiteGlobals.java
@@ -71,12 +71,11 @@ public abstract class TestSuiteGlobals {
                                         .header(SLUG, name)
                                         .when()
                                         .post(base);
-        if (res.getStatusCode() == 201) {
-            cleanupManager.setTestContainerUrl(containerUrl);
-            return containerUrl;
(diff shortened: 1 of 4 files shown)


src/main/java/org/fcrepo/spec/testsuite/TestSuiteGlobals.java
```diff
@@ -68,12 +72,33 @@ public static String containerTestSuite(final String baseurl, final String user,
                                         .when()
                                         .post(base);
         if (res.getStatusCode() == 201) {
+            cleanupManager.setTestContainerUrl(containerUrl);
             return containerUrl;
         } else {
             return base;
```
awoods: What do you think about throwing an exception if `res.getStatusCode != 201`?

bbpennel: That seems preferable to me, but wasn't sure if there was some expectation that this might fail for other implementations

awoods: Creating a top-level container seems like a reasonable assumption for any implementation.

PR #198, opened 2018-09-17
3.2.2-B Check all link headers for describes

whikloj:


src/main/java/org/fcrepo/spec/testsuite/crud/HttpGet.java
```diff
@@ -172,24 +173,15 @@ public void responseDescribesHeader(final String uri) {
...
-            // Loop multi-valued Link headers
-            for (final String linkValue : header.getValue().split(",")) {
-                final Link link = Link.valueOf(linkValue);
-                if (link.getRel().equals("describedby")) {
-                    description = link.getUri();
-                    break;
-                }
-            }
-        }
+
```
dbernstein: I believe AbstractTest.getLinksOfRelTypeAsUris() might be applicable here.

Issue #187, opened 2018-09-15
Tests needed: 4.3.3 HTTP POST


awoods:
https://fedora.info/2018/06/25/spec/#ldpcv-post

PR #200, opened 2018-09-17
CTS-187 - LDPCv 4.3.3

bbpennel:
Addresses https://github.com/fcrepo/Fedora-API-Test-Suite/issues/187

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpPost.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpPost.java
@@ -20,6 +20,7 @@ package org.fcrepo.spec.testsuite.versioning;
 import static org.fcrepo.spec.testsuite.Constants.MEMENTO_DATETIME_HEADER;
 import static org.fcrepo.spec.testsuite.Constants.MEMENTO_LINK_HEADER;
 import static org.testng.AssertJUnit.fail;
+import static org.testng.Assert.assertFalse;
 
 import java.net.URI;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpPost.java
```diff
@@ -48,6 +49,27 @@ public LdpcvHttpPost(final String username, final String password) {
...
+        final TestInfo info = setupTest("4.3", "ldpcvMayDisallowPost",
+                "Although an LDPCv is both a TimeMap and an LDPC, implementations MAY disallow POST requests.",
+                "https://fedora.info/2018/06/25/spec/#ldpcv-post",
+                ps);
+
+        // create the versioned resource
+        final Response createResponse = createVersionedResource(uri, info);
+
+        final URI timeMapURI = getTimeMapUri(createResponse);
+        final String allow = doOptions(timeMapURI.toString()).getHeader("Allow");
```
dbernstein: This test is failing when it should be succeeding.  Replace what you have at 69-70 with 

```
confirmPresenceOfHeaderValueInMultiValueHeader("Allow", "POST", doOptions(timeMapURI.toString()));
```

bbpennel: Section 4.3.3 states:
"Although an LDPCv is both a TimeMap and an LDPC, implementations MAY disallow POST requests"
This test is checking that POST may be disallowed. There is already https://github.com/fcrepo/Fedora-API-Test-Suite/blob/master/src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpOptions.java#L155 which tests that servers MAY support POST to LDPCvs. I'm not sure if there's a point to testing the negative, but it seems to be what this section is allowing.

We could drop this test if that would make more sense.

dbernstein: Ah - good point. My bad.   I think it can stay.

PR #201, opened 2018-09-17
Makes test ids consistent with the rest of project and fixes broken t…

dbernstein:
…est 5.0-C-2.

This test should now pass once FCREPO-2864 is merged.


src/main/java/org/fcrepo/spec/testsuite/authz/WebACGeneral.java
```diff
@@ -103,7 +103,7 @@ public void agentGroup(final String uri) {
         final String testContainerUri = createResource(uri, info.getId());
 
         //create agent-group list
-        final String groupListUri = testContainerUri + "/agent-group";
+        final String groupListUri = joinLocation(testContainerUri + "agent-group");
```
bbpennel: Should be a comma here instead of a plus

PR #203, opened 2018-09-17
CTS-196 - Added test for the possibility of allowing interaction models to change

bbpennel:
This test fails because it seems like we do not allow for interaction models to be changed in the community implementation. Or if we do I'm not sure what models are allowed to be changed between (versus adding new ones). The constraint document returned by the request isn't clarifying things too much:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-webapp/src/main/webapp/static/constraints/InteractionModelViolationException.rdf
and the behavior is not defined very well in the LDP spec (since its allowed as an extension, but not defined)

Alternatively, we could leave the stubbed out method there so we know that there is a test that needs to be filled in in the future when we have a case that can pass.

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/crud/HttpPut.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/crud/HttpPut.java
@@ -51,23 +51,46 @@ public class HttpPut extends AbstractTest {
     }
 
     /**
+     * 3.6-A
+     *
+     * @param uri The repository root URI
+     */
+    @Test(groups = {"MAY"})
+    @Parameters({"param1"})
+    public void httpPutChangeTypeAllowed(final String uri) {
+        final TestInfo info = setupTest("3.6-A", "httpPutChangeTypeAllowed",
+                "Implementations MAY allow the interaction model of an existing " +
+                        "resource to be changed by specification of a new LDP type in a " +
+                        "rel=\"type\" link in the HTTP Link header",
+                "https://fcrepo.github.io/fcrepo-specification/#http-put", ps);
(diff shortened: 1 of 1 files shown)


src/main/java/org/fcrepo/spec/testsuite/crud/HttpPut.java
```diff
@@ -90,7 +113,7 @@ public void httpPut(final String uri) {
      */
     @Test(groups = {"MUST"})
     @Parameters({"param1"})
-    public void httpPutpdateTriples(final String uri) {
+    public void httpPutUpdateTriples(final String uri) {
```
dbernstein: Line 117:  "httpPutpdateTriples" -> "httpPutUpdateTriples"

bbpennel: That is the change I made, it was a type before and now it is `httpPutUpdateTriples`

bbpennel: I generally get the impression this test suffered at the hand of multiple IDEs / merge issues, given the missing test, some of the formatting issues and this typo

bbpennel: Ah sorry misinterpreted this (waking up needed), i have updated line 117

awoods:
@bbpennel : The only allowed interaction model recastings are from supertype to subtype, as defined here:
https://www.w3.org/TR/ldp/images/ldpc-hierarchy.png

Issue #189, opened 2018-09-15
Tests needed: 5.7.1 LDP-RS


awoods:
https://fedora.info/2018/06/25/spec/#append-ldprs

PR #205, opened 2018-09-18
5.7.1 acl:Append on LDP-RS tests

### LINKED ISSUE #188: Tests needed: 5.7 Append Mode
https://fedora.info/2018/06/25/spec/#append-mode

whikloj:
Resolves #188 

The **MUST** PUT a child resource is contingent on the implementation allowing create on PUT. 

Not sure how we check for that so I left it in the MUST for now.

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/authz/WebACModes.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/authz/WebACModes.java
@@ -568,13 +568,6 @@ public class WebACModes extends AbstractAuthzTest {
         // verify failure.
         putResponse.then().statusCode(403);
 
-        // perform PUT to child resource as non-admin succeeds.
-        final Response putChildResponse =
-                doPutUnverified(resourceUri + "/child1", new Headers(new Header("Content-Type", "text/plain")),
-                        "test", false);
-        // verify successful
-        putChildResponse.then().statusCode(201);
-
         // perform PATCH which also deletes.
         final Response patchDelete = doPatchUnverified(resourceUri, new Headers(new Header("Content-type",
                 "application/sparql-update")),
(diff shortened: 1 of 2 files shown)


whikloj:
Actually this Resolves #189, #188 doesn't have anything testable in it.

src/main/java/org/fcrepo/spec/testsuite/authz/WebACModes.java
```diff
@@ -539,4 +539,90 @@ public void controlDisallowedPUT(final String uri) {
...
+
+        // perform PUT to resource as non-admin
+        final Response putResponse =
+                doPutUnverified(resourceUri, new Headers(new Header("Content-Type", "text/plain")),
+                        "test", false);
+        // verify failure.
+        putResponse.then().statusCode(403);
+
+        // perform PUT to child resource as non-admin succeeds.
+        final Response putChildResponse =
```
bbpennel: It may be best to break this case off into a separate test given "if the implementation supports creating resources using PUT" part, it is valid for this case to produce something other than a 201 if PUT is not supported

src/main/java/org/fcrepo/spec/testsuite/authz/WebACModes.java
```diff
@@ -539,4 +539,109 @@ public void controlDisallowedPUT(final String uri) {
...
+     *
+     * @param uri of base container of Fedora server
+     */
+    @Test(groups = { "MUST" })
+    @Parameters({ "param1" })
+    public void appendNotWriteLdpRsMust(final String uri) {
+        final TestInfo info = setupTest("5.7.1-A", "appendNotWriteLdpRsMust",
+                "When a client has acl:Append but not acl:Write for an LDP-RS they MUST " +
+                        "not DELETE, not PATCH that deletes triples, not PUT on the resource, and " +
+                        "must be able to PUT a child resource",
```
bbpennel: Might want to remove this part since its in a different test now

Issue #193, opened 2018-09-15
Tests needed: 5.9 Inheritance and Default ACLs


awoods:
https://fedora.info/2018/06/25/spec/#inheritance

dbernstein:
This one is done, pending a response from the spec editors on 
https://github.com/fcrepo/fcrepo-specification/issues/398

dbernstein:
The spec issue does not apply.  Waiting on https://github.com/fcrepo/Fedora-API-Test-Suite/pull/204 to be rebased so I can merge and rebase this branch and issue a PR.

PR #206, opened 2018-09-19
Implements test for 5.9 Inheritance and Default ACLs

dbernstein:
Resolves: https://github.com/fcrepo/Fedora-API-Test-Suite/issues/193

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/App.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/App.java
@@ -49,9 +49,9 @@ public class App {
      */
     public static void main(final String[] args) {
         final Options options = new Options();
-        final Option baseurl = new Option("b", "baseurl", true, "base url");
-        baseurl.setRequired(true);
-        options.addOption(baseurl);
+        final Option rootUrl = new Option("b", "rooturl", true, "The root URL of the repository");
+        rootUrl.setRequired(true);
+        options.addOption(rootUrl);
         final Option user = new Option("u", "user", true, "Username of user with basic user role");
         user.setRequired(true);
         options.addOption(user);
@@ -86,7 +86,7 @@ public class App {
             return;
         }
 
-        final String inputUrl = cmd.getOptionValue("baseurl");
+        final String inputUrl = cmd.getOptionValue("rooturl");
         final String inputUser = cmd.getOptionValue("user") == null ? "" : cmd.getOptionValue("user");
         final String inputPassword = cmd.getOptionValue("password") == null ? "" : cmd.getOptionValue("password");
         final String inputAdminUser = cmd.getOptionValue("admin-user") == null ? "" : cmd.getOptionValue("admin-user");
--- a/src/main/java/org/fcrepo/spec/testsuite/App.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/App.java
@@ -98,6 +98,7 @@ public class App {
 
         //Create the default container
         final Map<String, String> params = new HashMap<>();
+        params.put("param0", inputUrl);
         params.put("param1", TestSuiteGlobals.containerTestSuite(inputUrl, inputAdminUser, inputAdminPassword));
         params.put("param2", inputAdminUser);
         params.put("param3", inputAdminPassword);
--- a/src/main/java/org/fcrepo/spec/testsuite/authz/WebACDefaultACLs.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/authz/WebACDefaultACLs.java
@@ -1,0 +1,134 @@
+/*
+ * Licensed to DuraSpace under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * DuraSpace licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.spec.testsuite.authz;
+
+import static org.testng.AssertJUnit.assertEquals;
+
+import java.net.URI;
+
+import io.restassured.http.Header;
+import io.restassured.http.Headers;
+import io.restassured.response.Response;
+import org.fcrepo.spec.testsuite.TestInfo;
+import org.testng.annotations.Parameters;
+import org.testng.annotations.Test;
+
+/**
+ * @author dbernstein
+ * @since 2018-09-17
+ */
+public class WebACDefaultACLs extends AbstractAuthzTest {
+
+    /**
+     * Constructor
+     *
+     * @param adminUsername admin username
+     * @param adminPassword admin password
+     * @param username      username
+     * @param password      password
+     */
+    @Parameters({"param2", "param3", "param4", "param5"})
+    public WebACDefaultACLs(final String adminUsername, final String adminPassword, final String username,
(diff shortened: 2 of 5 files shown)


src/main/java/org/fcrepo/spec/testsuite/authz/WebACDefaultACLs.java
```diff
@@ -0,0 +1,134 @@
...
+        //create a grandchild  with no acl.
+        final Response grandChildPost = doPost(childUri, headers, simpleRDFBody);
+        final String grandChildUri = getLocation(grandChildPost);
+
+        // verify that the user can read and write to the grandparent
+        doGet(grandParentUri, false);
+        doPost(grandParentUri, headers, simpleRDFBody, false);
+
+        // verify that the user can read but not write to the child
+        doPostUnverified(childUri, headers, simpleRDFBody, false).then().statusCode(clientErrorRange());
```
bbpennel: It doesn't seem like we're totally consistent within the authz tests, but perhaps this check should be for 403 specifically? It seems that we want to ensure it is an authorization failure versus other bad requests, and that code should be pretty consistent across impls.

src/main/java/org/fcrepo/spec/testsuite/authz/WebACDefaultACLs.java
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * 5.9-B - In the case that the controlled resource is uncontained and has no ACL, or that there is no ACL at any
+     * point in the containment hierarchy of the controlled resource, then the server must supply a default ACL.
+     *
+     * @param uri of base container of Fedora server
+     */
+    @Test(groups = {"SHOULD"})
+    @Parameters({"param0"})
+    public void serverSuppliedDefaultAcl(final String uri) {
```
bbpennel: My interpretation is that this test is verifying "controlled resource is uncontained and has no ACL". Should we split out a second test to verify "that there is no ACL at any point in the containment hierarchy of the controlled resource"?

src/main/java/org/fcrepo/spec/testsuite/authz/WebACDefaultACLs.java
```diff
@@ -0,0 +1,134 @@
...
+    @Test(groups = {"SHOULD"})
+    @Parameters({"param0"})
+    public void defaultAclOnSameServer(final String uri) {
+        setupTest("5.9-D", "The default ACL resource should be located in the same server as the " +
+                           "controlled resource.",
+                  "https://fedora.info/2018/06/25/spec/#inheritance", ps);
+        //GET the default acl and verify that it is on the same server as the root uri.
+        final String aclUri = getAclLocation(uri);
+        assertEquals("The default ACL resource is not located in the same server as the controlled resource.",
+                     URI.create(uri).getHost(), URI.create(aclUri).getHost());
```
bbpennel: Is host comparison sufficient? Should this condition pass if the ACL is at a different port?

src/main/java/org/fcrepo/spec/testsuite/App.java
```diff
@@ -98,6 +98,7 @@ public static void main(final String[] args) {
 
         //Create the default container
         final Map<String, String> params = new HashMap<>();
+        params.put("param0", inputUrl);
```
bbpennel: I am wondering if we can assume that inputUrl is the root of the repository (that seems to be the intended use of this parameter). A user might specify a subcontainer to run the tests in rather than the root? In fcrepo-import-export we go through the trouble of seeking the root explicitly https://github.com/fcrepo4-labs/fcrepo-import-export/blob/master/src/main/java/org/fcrepo/importexport/exporter/Exporter.java#L442-L449

src/main/java/org/fcrepo/spec/testsuite/authz/WebACDefaultACLs.java
dbernstein: I suppose if by same server we mean same serving process, rather than host, I suppose we should check the host as well.

src/main/java/org/fcrepo/spec/testsuite/App.java
dbernstein: @bbpennel : I'm wondering if we should make the command line docs explicit - ie that the baseURL is the root, remove param0,  and change param1 to use the rootURI rather than create a separate test container since most if not all the tests create child test resources anyway.  What do you think?

src/main/java/org/fcrepo/spec/testsuite/authz/WebACDefaultACLs.java
dbernstein: good point.

src/main/java/org/fcrepo/spec/testsuite/authz/WebACDefaultACLs.java
dbernstein: I can certainly verify that the root resource is uncontained and has a default ACL.

dbernstein: Actually I don't think I can verify this.

dbernstein: I can check that the supplied root uri is in fact the root uri.

dbernstein: @bbpennel :  what did you have in mind for the second test?

src/main/java/org/fcrepo/spec/testsuite/App.java
bbpennel: I think the other purpose of the test container is to keep the test resources together in one place rather than cluttering up the root, so I'd rather not get rid of it. Declaring in the docs that baseURL has to be the root is probably fine, if that restriction is okay.

I feel a little weird about singling out the root solely for the purpose of getting an uncontained resource, but i'm not sure what else we could use. LDPCvs are uncontained (sort of), but they operate differently as far as ACLs go.

src/main/java/org/fcrepo/spec/testsuite/authz/WebACDefaultACLs.java
bbpennel: Wouldn't the "that there is no ACL at any point in the containment hierarchy of the controlled resource" just mean that you have a non-root container with no explicitly set ACL? The default acl on the root does make things a little confusing, but the child resource would be just as free of having an ACL assigned as the root..

src/main/java/org/fcrepo/spec/testsuite/App.java
dbernstein: Also, the lack of containment is implementation specific.  I believe it would be reasonable for others to be explicit about contains between the original and the ldpcv.

src/main/java/org/fcrepo/spec/testsuite/authz/WebACDefaultACLs.java
dbernstein: I suppose that would make sense.  I'll add it.

dbernstein: I suppose what complicates things further is that the default ACL is not specified.  So I'm not sure how I might test that.   I can create a child and verify that it's acl returns a 404.  But I would need to analyze the default acl to determine whether or not I should be able to read it with the non-admin user.

src/main/java/org/fcrepo/spec/testsuite/authz/WebACDefaultACLs.java
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * 5.9-B - In the case that the controlled resource is uncontained and has no ACL, or that there is no ACL at any
+     * point in the containment hierarchy of the controlled resource, then the server must supply a default ACL.
+     *
+     * @param uri of base container of Fedora server
+     */
+    @Test(groups = {"SHOULD"})
+    @Parameters({"param0"})
+    public void serverSuppliedDefaultAcl(final String uri) {
+        setupTest("5.9-C", "In the case that the controlled resource is uncontained and has no ACL, or " +
```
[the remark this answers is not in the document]
dbernstein: I will fix.

Issue #191, opened 2018-09-15
Tests needed: 5.7.3 LDP-NR


awoods:
https://fedora.info/2018/06/25/spec/#append-ldpnr

PR #208, opened 2018-09-19
5.7.3 acl:Append not acl:Write on LDP-NR

whikloj:
Resolves: #191 
Relies on #207 

This is built on top of an existing PR, so #207 should be merged first and then this one rebased.

This adds 3 tests for section 5.7.3 acl:Append on an LDP-NR

It has a path for the conceivable `PATCH`ing of the actual `LDP-NR` (instead of an associated `LDP-RS`) but that is an open question (https://github.com/fcrepo/fcrepo-specification/issues/400)

Currently it fails if you get there.

These tests all fail against the community implementation due to https://jira.duraspace.org/browse/FCREPO-2867

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/authz/WebACModes.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/authz/WebACModes.java
@@ -18,8 +18,6 @@
 package org.fcrepo.spec.testsuite.authz;
 
 import static org.fcrepo.spec.testsuite.Constants.APPLICATION_SPARQL_UPDATE;
-import static org.junit.Assert.fail;
-
 import io.restassured.http.Header;
 import io.restassured.http.Headers;
 import io.restassured.response.Response;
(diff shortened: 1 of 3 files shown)


src/main/java/org/fcrepo/spec/testsuite/authz/WebACModes.java
```diff
@@ -669,4 +670,150 @@ public void appendNotWriteLdpCMust(final String uri) {
...
+                    "prefix dc: <http://purl.org/dc/elements/1.1/> DELETE DATA { <> dc:title \"Some title\" .}",
+                    false);
+            // Verify failure.
+            patchDeleteData.then().statusCode(403);
+        }
+
+        if (patchLDPNR) {
+            final Response headResponse = doHead(resourceUri, false);
+            final String allowPatch = getHeaders(headResponse, "Allow-Patch").findFirst().map(Header::getValue).get();
+            // How are we testing patching an LDP-NR? Xdiff?
```
dbernstein: I'll add this one to the agenda for tomorrow's meething.

dbernstein: according to @awoods:  "PATCH is undefined for LDP-NR".  You can remove the final if block.

PR #210, opened 2018-09-19
Need to actually remove all testing of PATCH on LDP-NR

whikloj:
Follow-on to #208 and the decision that the API considers PATCH on LDP-NR to be undefined. ([Issue 5 here](https://wiki.duraspace.org/display/FF/2018-09-19+-+API+Spec+Meeting))

These tests can be removed


dbernstein:
@whikloj :  So you want to remove these tests because descriptions of LDP-NRs are LDP-RSs and therefore covered by other tests?

whikloj:
Yes the LDP-RS `MUST` not delete triples is handled [here](https://github.com/fcrepo/Fedora-API-Test-Suite/blob/master/src/main/java/org/fcrepo/spec/testsuite/authz/WebACModes.java#L571-L585) and `SHOULD` allow you to add triples is handled [here](https://github.com/fcrepo/Fedora-API-Test-Suite/blob/master/src/main/java/org/fcrepo/spec/testsuite/authz/WebACModes.java#L629-L644)

Issue #192, opened 2018-09-15
Tests needed: 5.8 Access To Class


awoods:
https://fedora.info/2018/06/25/spec/#access-to-class

PR #211, opened 2018-09-19
Adds tests for section 5.8 (accessToClass)

dbernstein:
Resolves https://github.com/fcrepo/Fedora-API-Test-Suite/issues/192

Test 5.8-B does not pass since we are not currently supporting inferencing in Fedora 5.0.0


mohideen:
@dbernstein  Looks good!

PR #213, opened 2018-09-20
Use config file

whikloj:
This is just to avoid having to type in the commands each time my terminal history is lost.

Also with the event testing we will need more arguments, this saves having to enter them all.

It can be made more strict if desired.


peichman-umd:
👍

Issue #46, opened 2018-03-20
Custom Want-Digest


ibr-jabrego:
The idea was to permit the user to use custom digests when running the testing. The application would permit the user to add two different paramaters -d1 and -d2. These paramaters would be digests used in the tests which use Want-Digest Headers in GET Requests. If not digests are included when running the application it would use md5 and sha as default. Let me know what you think. Thank you!

awoods:
In an effort to minimize client inputs in running these tests, I would suggest testing `sha`, `sha256`, and `md5`. If any of those algorithms are not supported, the proper error code should be returned.

PR #214, opened 2018-09-20
Adds sha-256 test and a test to confirm that a Want-Digest request

dbernstein:
header that does not contain a valid algorithm is rejected with a 400.
Resolves https://github.com/fcrepo/Fedora-API-Test-Suite/issues/46

The sha-256 header test (3.2.3-A-3) is expected to fail until https://jira.duraspace.org/browse/FCREPO-2876 is resolved.


peichman-umd:
Looks good to me. I don't have write permissions to this repo, so I can't actually merge it yet.

Issue #94, opened 2018-06-23
Test 3.7-C - is specific to the modeshape implementation, not API spec


zimeon:
The test 3.7-C, checking for illegal attempts to modify server managed triples:

https://github.com/fcrepo4-labs/Fedora-API-Test-Suite/blob/c30f47aa55c89ed1336cf4944aff104ce5e6bfa5/src/main/java/org/fcrepo/spec/testsuite/crud/HttpPatch.java#L141-L171

with SPARQL update code:

https://github.com/fcrepo4-labs/Fedora-API-Test-Suite/blob/c30f47aa55c89ed1336cf4944aff104ce5e6bfa5/src/main/java/org/fcrepo/spec/testsuite/crud/HttpPatch.java#L53-L57

is checking for rejection of a change to `fedora:lastModifiedBy` which is not defined in the API specification (or LDP of course).

Since there is no mechanism in the API specification to find out what special triples a server manages I think this test has to be removed from the test suite. I wonder whether there should be an extra collection of tests kept aside for the modeshape fedora 5?

awoods:
The essence of this requirement is that if the client performs a PATCH and the implementation is unable to satisfy that request, it should fail with a 409, [otherwise must return a 2xx response].

Would it make sense to have tests that exercise scenarios that fail against existing implementations, and if the response is 2xx, that passes the test as well?

zimeon:
Well, that would provide some test but would be very weak. A slightly stronger strategy I had thought of is the idea of creating an empty LDPRS, then trying to change the triples that appear when you GET it, on the assumption that these are server managed?

awoods:
Sounds like a plan.

dbernstein:
It seems the suggested approach will leave the functionality untested for implementations without server-managed triples (trellis).  @acoburn suggested simply trying to insert a triple disallowed by LDP such as <> rdf:type "a literal" since rdf:types must take a URI as an object. Any objections to this approach?

acoburn:
Just to clarify: LDP doesn't explicitly disallow the triple pattern `<> rdf:type "a literal"`, though that triple doesn't make much semantic sense. This is just an example of a triple pattern that an implementation might disallow (Trellis disallows that).

zimeon:
LDP does not require testing of RDF Schema validity so I don't see why a server is required to reject `<> rdf:type "A thingy!".`

zimeon:
[...]

PR #216, opened 2018-09-21
Ensures that 3.7-C,D,and E are not implementation dependent.

dbernstein:
Resolves: https://github.com/fcrepo/Fedora-API-Test-Suite/issues/94


src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java
```diff
@@ -549,8 +550,13 @@ protected void confirmPresenceOfLinkValue(final Link linkValue, final Response r
 
     protected void confirmPresenceOfConstrainedByLink(final Response response) {
         final String contrainedByUri = "http://www.w3.org/ns/ldp#constrainedBy";
-        assertTrue("Response does not contain link of rel type = " + contrainedByUri,
-                   getLinksOfRelType(response, contrainedByUri).count() > 0);
+        final Optional<URI> link = getLinksOfRelTypeAsUris(response, contrainedByUri).findAny();
+        if (!link.isPresent()) {
+            fail("Response does not contain a link of rel type = " + contrainedByUri);
```
bbpennel: might as well fix the typo while we're here, `constrainedByUri`

src/main/java/org/fcrepo/spec/testsuite/crud/HttpPatch.java
```diff
@@ -122,12 +128,49 @@ public void serverManagedPropertiesModification(final String uri) {
...
+        // server-managed triple.
         final Response resource = createBasicContainer(uri, info);
-        final String locationHeader = getLocation(resource);
+        final String resourceUri = getLocation(resource);
+
+        //parse the resource's body
+        final Model model = createDefaultModel();
+        model.read(doGet(resourceUri, new Header("Accept", "text/turtle")).getBody().asInputStream(), "",
+                   "text/turtle");
+        final Dataset dataset = new DatasetImpl(model);
```
bbpennel: In this case I don't think you need to go through a Dataset, you should be able to operate on the model:
```
        final StmtIterator it = model.listStatements();
        boolean atLeastOneUpdateFailed = false;

        while (it.hasNext()) {
            //for each triple
            final Statement oldValue = it.next();
```

src/main/java/org/fcrepo/spec/testsuite/crud/HttpPatch.java
```diff
@@ -122,12 +128,49 @@ public void serverManagedPropertiesModification(final String uri) {
...
+        final DatasetGraph graph = dataset.asDatasetGraph();
+        final Iterator<Quad> it = graph.find();
+        boolean atLeastOneUpdateFailed = false;
+
+        while (it.hasNext()) {
+            //for each triple
+            final Triple oldValue = it.next().asTriple();
+            final String newObject = "newValue:" + System.currentTimeMillis();
+            //attempt to update the resource with that triple
+            if (patchReturned4xx(resourceUri, oldValue.getPredicate().getURI(),
```
bbpennel: If another implementation requires certain fields to have certain types of object values (lastModified must be a date, or ldp:contains must be a URI) this would fail due to converting everything to a string literal instead of the server managed triple restriction. On the other hand, the language of this section just says "request is received that attempts to modify statements to a resource that a server disallows", so maybe this falls into the same boat.

src/main/java/org/fcrepo/spec/testsuite/crud/HttpPatch.java
```diff
@@ -143,13 +186,38 @@ public void statementNotPersistedResponseBody(final String uri) {
...
-        doPatchUnverified(locationHeader, headers, serverProps)
-                .then()
-                .statusCode(clientErrorRange())
-                .body(containsString("lastModified"));
+        final String resourceUri = getLocation(resource);
+
+        final Model model = createDefaultModel();
+        model.read(doGet(resourceUri, new Header("Accept", "text/turtle")).getBody().asInputStream(), "",
+                   "text/turtle");
+        final Dataset dataset = new DatasetImpl(model);
```
bbpennel: Can skip the Dataset here as well

src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java
bbpennel: I'm also not clear why we now need to verify that the constrainedby link is resolvable? Not that its a bad thing to check

src/main/java/org/fcrepo/spec/testsuite/crud/HttpPatch.java
```diff
@@ -166,13 +234,39 @@ public void statementNotPersistedConstrainedBy(final String uri) {
...
-                .then()
-                .statusCode(clientErrorRange())
-                .header("Link", containsString("constrainedBy"));
+        final String resourceUri = getLocation(resource);
+
+        //parse the resource's body
+        final Model model = createDefaultModel();
+        model.read(doGet(resourceUri, new Header("Accept", "text/turtle")).getBody().asInputStream(), "",
+                   "text/turtle");
+        final Dataset dataset = new DatasetImpl(model);
```
bbpennel: Can skip the Dataset here as well

src/main/java/org/fcrepo/spec/testsuite/crud/HttpPatch.java
```diff
@@ -122,12 +128,49 @@ public void serverManagedPropertiesModification(final String uri) {
                                         + "4.2.4.1), the server must fail the request by responding with a 4xx range"
                                         + " status code (e.g. 409 Conflict).",
                                         "https://fcrepo.github.io/fcrepo-specification/#http-patch", ps);
+
+        //This test assumes that creating an empty container will result in RDF that contains at least one
+        // server-managed triple.
```
acoburn: Where does the Fedora specification say anything about the creation of an empty container resulting in at least one server-managed triple? In fact, last I checked, the specification has nothing to say about "server managed" triples. I believe that "server-managed triples" is a Fedora4 invention. I can assure you that Trellis allows a client to create an empty resource, resulting in....an empty resource.

src/main/java/org/fcrepo/spec/testsuite/crud/HttpPatch.java
acoburn: Yes, Trellis has restrictions on certain values (e.g. neither `ldp:membershipResource` nor `rdf:type` can refer to a `Literal`). I think you'd be best served by looking at the restrictions that LDP already defines and testing those.

src/main/java/org/fcrepo/spec/testsuite/crud/HttpPatch.java
acoburn: Easy:

1. try setting `ldp:contains` to some arbitrary value.
2. try setting `ldp:insertedContentRelation` on a basic container (not allowed under LDP)
3. try setting a semantically invalid statement: `<> a "literal value"`
4. try setting mutually exclusive properties on a direct container: i.e. both `ldp:hasMemberRelation` and `ldp:isMemberOfRelation` (not allowed under LDP)
5. try setting `ldp:membershipResource` to some out-of-domain resource (I'm not sure what that would even mean on an LDP server).

Surely some (or all) of those would fail on a good LDP implementation.

src/main/java/org/fcrepo/spec/testsuite/crud/HttpPatch.java
dbernstein: Fair enough.

src/main/java/org/fcrepo/spec/testsuite/crud/HttpPatch.java
dbernstein: Thanks.

src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java
dbernstein: @bbpennel : Re resolution:  So is it possible that a constrainedBy URI would not be resolvable? In other words, of what use would an unresolvable be?

src/main/java/org/fcrepo/spec/testsuite/App.java
```diff
@@ -130,6 +133,8 @@ public static void main(final String[] args) {
         options.addOption(new Option("k", BROKER_URL_PARAM, true, "The URL of the JMS broker."));
         options.addOption(new Option("q", QUEUE_NAME_PARAM, true, "Queue name for events (if applicable)"));
         options.addOption(new Option("t", TOPIC_NAME_PARAM, true, "Topic name for events (if applicable)"));
+        options.addOption(new Option("c", CONSTRAINT_ERROR_GENERATOR_PARAM, true,
```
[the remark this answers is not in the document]
dbernstein: Updated.

src/main/java/org/fcrepo/spec/testsuite/crud/HttpPatch.java
```diff
@@ -232,8 +244,8 @@ public void disallowPatchContainmentTriples(final String uri) {
 
         final Headers headers = new Headers(new Header("Content-Type", APPLICATION_SPARQL_UPDATE));
         doPatchUnverified(locationHeader, headers, updateContainmentTriples)
-                .then()
```
awoods: Please revert reformatting.

src/main/java/org/fcrepo/spec/testsuite/crud/HttpPatch.java
```diff
@@ -255,7 +267,7 @@ public void disallowChangeResourceType(final String uri) {
 
         final Headers headers = new Headers(new Header("Content-Type", APPLICATION_SPARQL_UPDATE));
         doPatchUnverified(locationHeader, headers, resourceType)
-                .then()
```
awoods: Please revert reformatting.

src/main/java/org/fcrepo/spec/testsuite/crud/HttpPatch.java
```diff
@@ -35,40 +38,40 @@
...
+                                   + "Add {"
+                                   + " <#> dcterms:description \"Patch LDP Updated Description\" ;"
+                                   + "} .";
     private final String resourceType = "PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>"
-                                 + " PREFIX ldp: <http://www.w3.org/ns/ldp#>"
-                                 + " INSERT {"
-                                 + " <> rdf:type ldp:NonRDFSource ."
-                                 + "}"
-                                 + " WHERE { }";
+                                        + " PREFIX ldp: <http://www.w3.org/ns/ldp#>"
```
awoods: Please revert reformatting.

src/main/java/org/fcrepo/spec/testsuite/crud/HttpPatch.java
```diff
@@ -35,40 +38,40 @@
...
+                                        + " INSERT {"
+                                        + " <> rdf:type ldp:NonRDFSource ."
+                                        + "}"
+                                        + " WHERE { }";
     private final String updateContainmentTriples = "PREFIX ldp: <http://www.w3.org/ns/ldp#>\n"
-                                             + " INSERT {   \n"
-                                             + "  <> ldp:contains \"some-url\" .\n"
-                                             + "}\n"
-                                             + " WHERE { }";
+                                                    + " INSERT {   \n"
```
awoods: Please revert reformatting.

src/main/java/org/fcrepo/spec/testsuite/crud/HttpPatch.java
```diff
@@ -35,40 +38,40 @@
 public class HttpPatch extends AbstractTest {
 
     private final String body = "PREFIX dcterms: <http://purl.org/dc/terms/>"
-                         + " INSERT {"
```
awoods: Please revert reformatting.

src/main/java/org/fcrepo/spec/testsuite/crud/HttpPatch.java
```diff
@@ -35,40 +38,40 @@
...
-                         + " <> dcterms:description \"Patch Updated Description\" ."
-                         + "}"
-                         + " WHERE { }";
-    private final String ldpatch = "@prefix dcterms: <http://purl.org/dc/terms/>"
-                            + "Add {"
```
awoods: Please revert reformatting.

Issue #195, opened 2018-09-15
Tests needed: 6.2 Notification Serialization


awoods:
https://fedora.info/2018/06/25/spec/#notification-serialization

PR #220, opened 2018-09-21
6.1 Notification events

### LINKED ISSUE #194: Tests needed: 6.1 Notification Events
https://fedora.info/2018/06/25/spec/#notification-events

whikloj:
Resolves #194 
Partially #195 

The changes to `App.java` are just to replace all the instances of typed options to variables because I extend the JMS ones in the the test.

Test 6.1 just counts the messages, I think that is adequate.

Test 6.2 tests for the MUST statements about format, but only for a POST. It could be extended for other actions.

This does **NOT** test the `SHOULD` requirements of [6.2](https://fedora.info/2018/06/25/spec/#notification-serialization)


src/main/java/org/fcrepo/spec/testsuite/event/NotificationTest.java
```diff
@@ -0,0 +1,187 @@
...
+        // Get a new consumer.
+        final MessageConsumer consumer = getConsumer();
+        // Assign a message bank to capture the messages.
+        consumer.setMessageListener(new MessageBank());
+        // Start listening to the broker.
+        connection.start();
+        // Do your actions.
+        final Response response = createBasicContainer(uri, info);
+        final String location = getLocation(response);
+        Thread.sleep(500);
```
acoburn: You'll be happier if you use [awaitility](https://github.com/awaitility/awaitility) instead of `Thread.sleep`.

acoburn:
@whikloj this is how I test events in Trellis: https://github.com/trellis-ldp/trellis/blob/master/components/test/src/main/java/org/trellisldp/test/EventTests.java. This class is entirely independent of message broker technology. The broker technology gets added in at the point where a [particular application is tested](https://github.com/trellis-ldp/trellis/blob/master/components/app-triplestore/src/test/java/org/trellisldp/app/triplestore/TrellisEventTest.java) and all the data is simply passed to the underlying interfaces which is where the testing happens. This means that I can use JMS, AMQP or any other new-fangled messaging technology without changing any of the underlying tests.

whikloj:
I know your suggestions are good, I'm just not sure I understand them or how to wire them together and it took me forever to even get this working. My brain is not what it once was. 😉 

Though I might be able to parse JSON-LD because I did have a concern around not covering the 
```
@context: {
   @vocab: "https://www.w3.org/ns/activitystreams"
   ...
```

format

whikloj:
But I think I can figure out `awaitilly`

acoburn:
As a point of reference for using `awaitility`, you can look at the JMS module in the Fedora4 codebase:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-jms/src/test/java/org/fcrepo/integration/jms/observer/AbstractJmsIT.java#L191-L197. Just add this as a dependency:

https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-jms/pom.xml#L133-L136

whikloj:
Ok, I'm stopping. This is equivalent to what I left off with on Friday (maybe a little more). But probably easier to work with then json-pathing everything.

I left 2 `TODO`s for the other `SHOULD` tests and after reviewing all the tests @acoburn has in Trellis I'm sure we should be at least doing some tests for each of the ldp:Container types.

We could possibly merge 6.1 and 6.2-A tests and for each `POST`, `PUT`, `PATCH`, `DELETE` check that the activity type matches an activity streams one.

I literally put `AS:Create` and `AS:Update` but as the spec does not specify those, we probably should remove those and make it use a statement iterator and look at each type to ensure at least one has the AS namespace.

whikloj:
Pinging @fcrepo/committers, @acoburn if you'd have a look too I'd appreciate all the eyes I can get.

src/main/java/org/fcrepo/spec/testsuite/event/NotificationTest.java
```diff
@@ -0,0 +1,274 @@
...
+        final Response response = createBasicContainer(uri, info);
+        final String location = getLocation(response);
+        // Get the message bank back.
+        final MessageBank listener = (MessageBank) consumer.getMessageListener();
+        await().atMost(TEN_SECONDS).until(() -> listener.stream()
+                .count() == 2);
+
+        final Resource locResource = ResourceFactory.createResource(location);
+        final List<Message> message = listener.stream().collect(toList());
+        for (final Message m : message) {
```
acoburn:
```java
listener.stream().forEach(m -> {
    ...
 });
```

src/main/java/org/fcrepo/spec/testsuite/event/NotificationTest.java
```diff
@@ -0,0 +1,274 @@
...
+            if (m instanceof TextMessage) {
+                final String body = ((TextMessage) m).getText();
+                final InputStream is = new ByteArrayInputStream(body.getBytes("UTF-8"));
+                final Model model = ModelFactory.createDefaultModel();
+
+                model.read(is, location, "JSON-LD");
+                if (model.contains(locResource, RdfType)) {
```
acoburn: You're not checking the agent value here. Should you be?

acoburn: Should you check that each event has its own unique identifier? (This is different than the object IRI)

whikloj: This test is all the `MUST` requirements so as Agent is a `SHOULD` I was putting those tests below in 6.2-B. As for unique identifier for each event,  I don't think there is that requirement in the specification so while it makes sense I don't know if we need to test for it.

whikloj:
I also agree that this should be expanded with tests against various types of resources and operations. Sadly that will have to be the next person.

config.yml.dist
```diff
@@ -1,6 +1,9 @@
 default:
-  rooturl: http://localhost:8088/rest
+  rooturl: http://localhost:8080/rest
   admin-user: fedoraAdmin
   admin-password: fedoraAdmin
   user: testuser
   password: testpass
+  broker-url: tcp://127.0.0.1:61616
```
dbernstein: @whikloj : you also need to add the broker-url,topic, and/or queue name to the command line parameters since the config file is optional.  Without -c The application hangs.

whikloj: Do you want new CLI options for these 3, or just have it verify their existence in the config file?

dbernstein: I would say for consistency add them to CLI options and make the tool fail if they are not configured in either place.

Issue #224, opened 2018-10-03
Changing an LDP-R to LDP-Rv with put is not being tested properly


dbernstein:
4.0-B is creating an LDP-Rv with a PUT using an OriginalResource link header.  It should be creating an LDP-R first and then issuing a PUT with a link header in order to _update_ the resource.

PR #225, opened 2018-10-03
Changing an LDP-R to LDP-Rv with put is now being tested properly.

dbernstein:
Resolves: https://github.com/fcrepo/Fedora-API-Test-Suite/issues/224

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/versioning/ResourceVersioning.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/versioning/ResourceVersioning.java
@@ -77,25 +77,27 @@ public class ResourceVersioning extends AbstractVersioningTest {
      */
     @Test(groups = { "MUST" })
     @Parameters({ "param1" })
-    public void putLdprWithType(final String uri) {
+    public void convertToLDPRViaPutWithType(final String uri) {
         final TestInfo info = setupTest("4.0-B",
-                "When an LDPR is created with a rel=\"type\" link in the Link " +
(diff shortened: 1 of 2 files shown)


src/main/java/org/fcrepo/spec/testsuite/versioning/ResourceVersioning.java
```diff
@@ -77,25 +77,27 @@ public void postLdprWithType(final String uri) {
      */
     @Test(groups = { "MUST" })
```
awoods: This probably goes to "MAY"

src/main/java/org/fcrepo/spec/testsuite/versioning/ResourceVersioning.java
```diff
@@ -77,25 +77,27 @@ public void postLdprWithType(final String uri) {
...
     @Parameters({ "param1" })
-    public void putLdprWithType(final String uri) {
+    public void convertToLDPRViaPutWithType(final String uri) {
         final TestInfo info = setupTest("4.0-B",
-                "When an LDPR is created with a rel=\"type\" link in the Link " +
-                        "header specifying type http://mementoweb.org/ns#OriginalResource " +
-                        "to indicate versioning, it MUST be created as an LDPRv and a version " +
-                        "container (LDPCv) MUST be created to contain Memento resources",
+                "Implementations may allow a subsequent PUT request with a rel=\"type\" link in the Link header " +
```
awoods: For emphasis, let's all-caps `may` as well as `must` on line 84.

PR #239, opened 2018-10-11
Fedora Test Suite Test 3.5.1-B fails

robyj:
- Multiple LINK headers are received, only last one is recognized.
  Test has been modified to make the headers a string and find exact
  string in that string

Resolves https://jira.duraspace.org/browse/FCREPO-2918


awoods:
Superseded by: https://github.com/fcrepo/Fedora-API-Test-Suite/pull/249

Issue #232, opened 2018-10-11
Link parsing


acoburn:
The [Memento specification](https://tools.ietf.org/html/rfc7089) describes a format for Link headers:

```http
Link: <http://a.example.org/>; rel="original timegate",
    <http://a.example.org/?version=all&style=timemap>
      ; rel="timemap"; type="application/link-format"
      ; from="Tue, 15 Sep 2000 11:28:26 GMT"
      ; until="Wed, 20 Jan 2010 09:34:33 GMT"
```

The `from` and `until` parameters are not required, but they are certainly not unexpected. However, if a server generates Memento links in this format, this test suite fails, being unable to parse the link headers. The reason for this has to do with the presence of a comma in the date strings in `from` and `until`.

Note this [line](https://github.com/fcrepo/Fedora-API-Test-Suite/blob/master/src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java#L539) where a link header is simply split on commas regardless of where that comma may appear.

As such, most link-related tests (of which there are many) will fail if the origin server generates memento links according to the memento specification above. It is worth noting that commas are entirely legit if they appear inside quotes in link parameter values.

PR #250, opened 2018-10-15
Improve parsing of Link headers

awoods:
In combination with: https://github.com/fcrepo/Fedora-API-Test-Suite/pull/249 ...

Resolves: https://github.com/fcrepo/Fedora-API-Test-Suite/issues/232

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/crud/Container.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/crud/Container.java
@@ -17,8 +17,6 @@
  */
 package org.fcrepo.spec.testsuite.crud;
 
-import java.util.List;
-
 import io.restassured.http.Header;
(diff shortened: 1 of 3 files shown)


acoburn:
Thanks. I'll have some more issues along these lines, once this and the other open PRs are merged.

You may want to look more closely at the `Link` [comparison code](https://github.com/fcrepo/Fedora-API-Test-Suite/blob/master/src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java#L531-L534) and consider what happens if an incoming Link header contains more than just a `rel='type'` property (e.g. as defined in [these constants](https://github.com/fcrepo/Fedora-API-Test-Suite/blob/master/src/main/java/org/fcrepo/spec/testsuite/Constants.java#L58-L64) )

PR #266, opened 2018-10-25
Fixes test 5.0-V

dbernstein:
Resolves: https://jira.duraspace.org/browse/FCREPO-2911


src/main/resources/acls/agent-group.ttl
```diff
@@ -1,5 +1,5 @@
 @prefix    acl:  <http://www.w3.org/ns/auth/acl#>.
 @prefix  vcard:  <http://www.w3.org/2006/vcard/ns#>.
 <>  a                vcard:Group;
-    vcard:hasMember  "${user}";
+    vcard:hasMember  "${user}" .
```
awoods: This file does not seem related to the failing test.

dbernstein: oh - you're right.  I'm not sure how that got in there.  I'l remove.

Issue #274, opened 2018-10-25
crud.HttpPut tests must throw SkipException when optional features are not supported


dbernstein:
putCreateRDFSource()
putCreateNonRDFSource()

PR #280, opened 2018-10-30
Skip non-supported crud.HttpPut tests

awoods:
Resolves: https://github.com/fcrepo/Fedora-API-Test-Suite/issues/274

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/crud/HttpPut.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/crud/HttpPut.java
@@ -357,7 +357,7 @@ public class HttpPut extends AbstractTest {
     /**
      * 3.6.3-A
      */
-    @Test(groups = {"MUST"})
+    @Test(groups = {"MAY"})
     public void putCreateRDFSource() {
         final TestInfo info = setupTest("3.6.3-A",
                 "Implementations may accept HTTP PUT to create resources",
@@ -373,7 +373,7 @@ public class HttpPut extends AbstractTest {
         final Response putResponse = doPutUnverified(childURL);
         final int statusCode = putResponse.statusCode();
 
-        if (statusCode >= 400 && statusCode < 500) {
+        if (clientErrorRange().matches(statusCode)) {
             throw new SkipException("PUT to create not supported");
 
         } else {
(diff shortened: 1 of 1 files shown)


src/main/java/org/fcrepo/spec/testsuite/crud/HttpPut.java
```diff
@@ -355,36 +357,60 @@ public void nonRDFSourcePutDigestResponseHeaderVerification() {
     /**
      * 3.6.3-A
      */
-    @Test(groups = {"MAY"})
+    @Test(groups = {"MUST"})
```
dbernstein: Why is this a must?  The spec says may.

dbernstein: Looking at the spec I see the must in there:  I would suggest adding the following to the test description: 
```
On creation of an LDP-NR with HTTP PUT, implementations must create an associated LDP-RS describing that LDP-NR in the same way that it would when 3.5.1 Creating LDP-NRs with HTTP POST.
```

src/main/java/org/fcrepo/spec/testsuite/crud/HttpPut.java
```diff
@@ -355,36 +357,60 @@ public void nonRDFSourcePutDigestResponseHeaderVerification() {
...
-                .then()
-                .statusCode(201);
+        // Is PUT to-create supported?
+        final String slug = Integer.toString(new Random().nextInt());
+        final String childURL = joinLocation(containerLocation, slug);
+
+        final Response putResponse = doPutUnverified(childURL);
+        final int statusCode = putResponse.statusCode();
+
+        if (statusCode >= 400 && statusCode < 500) {
```
dbernstein: May I suggest using:  `clientErrorRange().matches(statusCode)`

src/main/java/org/fcrepo/spec/testsuite/crud/HttpPut.java
```diff
@@ -355,36 +357,60 @@ public void nonRDFSourcePutDigestResponseHeaderVerification() {
...
+        final String childURL = joinLocation(containerLocation, slug);
 
-        doPutUnverified(location, headers)
-                .then()
-                .statusCode(201);
+        final Headers headers = new Headers(new Header("Link", Constants.NON_RDF_SOURCE_LINK_HEADER));
+        final Response putResponse = doPutUnverified(childURL, headers);
+        final int statusCode = putResponse.statusCode();
+
+        if (statusCode >= 400 && statusCode < 500) {
```
dbernstein: May I suggest using: clientErrorRange().matches(statusCode)?

src/main/java/org/fcrepo/spec/testsuite/crud/HttpPut.java
awoods: Changing back to `MAY`.

PR #282, opened 2018-10-30
Skip non-supported crud.HttpDelete tests

awoods:
Resolves: https://github.com/fcrepo/Fedora-API-Test-Suite/issues/271


src/main/java/org/fcrepo/spec/testsuite/crud/HttpDelete.java
```diff
@@ -227,6 +226,14 @@ public void httpDeleteStatusCheckTwo() {
         ps.append("Headers:\tAccept=*/*\n");
         ps.append("Body:\n");
 
+        // Options to rootres
+        final Response responseOptions = doOptions(locationHeader);
+
+        // Is DELETE supported?
+        if (getHeaders(responseOptions, "Allow").noneMatch(h -> h.getValue().contains("DELETE"))) {
```
dbernstein: Would you consider refactoring the DELETE support check into it's own function?

Issue #276, opened 2018-10-29
New tests: 3.1.2 LDP Direct Containers


awoods:
New tests should be implemented for the new section on LDP Direct Containers:
https://fcrepo.github.io/fcrepo-specification/#ldpdc

PR #287, opened 2018-10-31
New tests: MAY support Direct and Indirect containers

dbernstein:
Resolves: https://github.com/fcrepo/Fedora-API-Test-Suite/issues/276


src/main/java/org/fcrepo/spec/testsuite/crud/Container.java
```diff
@@ -35,16 +36,53 @@
...
+    public void createLDPIndirectContainer() {
+        final TestInfo info = setupTest("3.1.1-A-3",
+                                        "Implementations may support the creation and management of [LDP] Indirect " +
+                                        "Containers",
+                                        "https://fcrepo.github.io/fcrepo-specification/#ldpc", ps);
+        skipIfIndirectContainersNotSupported();
+    }
+
+    private void skipIfDirectContainersNotSupported() {
+        if (clientErrorRange().matches(createDirectContainerUnverifed(uri, "").getStatusCode())) {
```
awoods: For clarity, please add a small dummy body, such as `"some-rdf"`

src/main/java/org/fcrepo/spec/testsuite/crud/Container.java
```diff
@@ -35,16 +36,53 @@
...
+            throw new SkipException("This implementation does not support DirectContainers");
+        }
+    }
+
+    private void skipIfIndirectContainersNotSupported() {
+        if (clientErrorRange().matches(createIndirectContainerUnverifed(uri, "").getStatusCode())) {
```
awoods: For clarity, please add a small dummy body, such as `"some-rdf"`

src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java
```diff
@@ -205,6 +205,13 @@ protected Response createDirectContainerUnverifed(final String uri, final String
         return doPostUnverified(uri, headers, body);
     }
 
+    protected Response createIndirectContainerUnverifed(final String uri, final String body) {
```
awoods: It is not clear that a `body` parameter is needed here, since the only usage of this method passes in `""`. Maybe create a dummy body within this method?

src/main/java/org/fcrepo/spec/testsuite/crud/Container.java
dbernstein: 👍

src/main/java/org/fcrepo/spec/testsuite/crud/Container.java
dbernstein: 👍

src/main/java/org/fcrepo/spec/testsuite/AbstractTest.java
dbernstein: 👍

PR #330, opened 2019-08-06
Direct container skip

gregjan:
This code primarily separates direct and indirect container support from other related requirements, including the ldp:constrainedBy header requirement. It ensured that some additional tests are skipped when direct or indirect containers are not supported (Prefer membership). It creates a separate set of tests for the ldp:constrainedBy Link header.

Note: This will require rebase after/if other PRs are accepted.

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/crud/Container.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/crud/Container.java
@@ -138,12 +138,23 @@ public class Container extends AbstractTest {
         final Response container = createBasicContainer(getLocation(base), "container");
         final Response containerChild = createBasicContainer(getLocation(container), "child");
 
-        final Response direct = createDirectContainer(
-                getLocation(base), DIRECT_CONTAINER_BODY.replace("%membershipResource%", getLocation(container)));
-        final Response directMember = createBasicContainer(getLocation(direct), "member");
+        // Skips adding a direct container and the triple check, if not supported
+        boolean directContainerTripleCheck = true;
+        Response directMember = null;
+        try {
+            skipIfDirectContainersNotSupported();
+            final Response direct = createDirectContainer(
+                    getLocation(base), DIRECT_CONTAINER_BODY.replace("%membershipResource%", getLocation(container)));
+            directMember = createBasicContainer(getLocation(direct), "member");
+        } catch (SkipException e) { // DC not supported
+            directContainerTripleCheck = false;
+        }
 
+        // Only omit, not include, can guarantee the omission of specified triples.
         final Response resP = doGet(getLocation(container),
-                new Header("Prefer", "return=representation; include=\"http://www.w3.org/ns/ldp#PreferContainment\""));
+                new Header("Prefer", "return=representation; " +
+                                     "omit=\"http://www.w3.org/ns/ldp#PreferMembership " +
+                                     "http://www.w3.org/ns/ldp#PreferMinimalContainer\""));
 
         ps.append(resP.getStatusLine()).append("\n");
         final Headers headers = resP.getHeaders();
@@ -158,9 +169,15 @@ public class Container extends AbstractTest {
         confirmPresenceOrAbsenceOfTripleInResponse(resP, getLocation(container), LDP_CONTAINS_PREDICATE,
                                                    getLocation(containerChild), true);
 
-        // Verify absence of unexpected triple in response body
-        confirmPresenceOrAbsenceOfTripleInResponse(resP, getLocation(container), LDP_CONTAINS_PREDICATE,
-                                                   getLocation(directMember), false);
+        if (directContainerTripleCheck) {
+            // Verify absence of unexpected triple in response body
+            confirmPresenceOrAbsenceOfTripleInResponse(resP, getLocation(container), LDP_CONTAINS_PREDICATE,
+                                                       getLocation(directMember), false);
+        }
+
+        // Check for presence of the dcterms:title triple.
+        confirmPresenceOrAbsenceOfPredicateInResponse(resP, getLocation(container), "http://purl.org/dc/terms/title",
+                false);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


gregjan:
There is a lot here and probably several commits that are worthy of discussion. To summarize, several tests resulted in errors for reasons particular to Trellis, but not out of line with the Fedora spec (IMO). So in those cases I have narrowed the tests to focus on the specification. In other cases I have expanded the test code because Trellis requires a more complete RDF body for direct and indirect containers. I also added a less prescriptive Graph assertion method, for testing triple patterns instead of particular SPOs.

gregjan:
FYI, latest annotated results for Trellis w/PostgreSQL after these adjustments: http://drastic-testbed.umd.edu/static/Fedora-reports/trellis-ext-db/report/testsuite-execution-report.html

src/main/java/org/fcrepo/spec/testsuite/crud/Container.java
```diff
@@ -158,9 +170,15 @@ public void ldpcContainmentTriples() {
...
-        confirmPresenceOrAbsenceOfTripleInResponse(resP, getLocation(container), LDP_CONTAINS_PREDICATE,
-                                                   getLocation(directMember), false);
+        if (directContainerTripleCheck) {
+            // Verify absence of unexpected triple in response body
+            confirmPresenceOrAbsenceOfTripleInResponse(resP, getLocation(container), LDP_CONTAINS_PREDICATE,
+                                                       getLocation(directMember), false);
+        }
+
+        // Check for presence of the dcterms:title triple.
+        confirmPresenceOrAbsenceOfPredicateInResponse(resP, getLocation(container), "http://purl.org/dc/terms/title",
```
bseeger: In my testing running against fcrepo4 version 5.1.0, it fails here. This is because the test suite creates the container `container` with the [BASIC_CONTAINER_BODY](https://github.com/fcrepo/Fedora-API-Test-Suite/blob/master/src/main/java/org/fcrepo/spec/testsuite/Constants.java#L32), which has a dc:title in the body.  I'm not clear on what feature you're testing with this check here.

bseeger: Altogether possible I'm missing some variable in the testing here, though.

gregjan: Sounds like Fedora has the right semantics to pass my predicate check, but I'll have a look to see if I can explain that lack of namespace expansion. I thought that was automatic, but it may not be. Can you determine if the Fedora response includes the dc namespace declaration, as usual?

gregjan: Oh, nevermind. That confirmPresenceOrAbsenceOfPredicateInResponse call is there to make sure that the response *only* includes containment triples and nothing else, which is what I expected by using "Prefer ... omit=...PreferMinimalContainer". It sounds like Fedora is not honoring the omit instruction. I'm going to check back in LDP specs to make sure I have the semantics right.

bseeger: Ah, I'll look into that as well from the fcrepo side of things.

gregjan: Okay, so this section of LDP defines the set of PreferMinimalContainer triples as the set of triples that would be present if the container were empty, i.e. dc:title. So dc:title should not be there if I ask the server to omit PreferMinimalContainer.

gregjan: link: https://www.w3.org/2012/ldp/hg/ldp.html#dfn-minimal-container-triples

bseeger: Actually, I wonder if this is a configuration of fcrepo issue on my end.  I'm going to try one thing and get back to you.

bseeger: I need to look into this a little more (later), because my read of that is that it would include any triples that were not containment or membership triples, so wouldn't dc:title be present still?  Also, I think the PreferMinimalContainer needs to be in the `include` part of the prefer header - or rather fcrepo is expecting it to be in the `include` and not `omit` (this could be wrong - not sure).  Although, honestly, this is all a bit fuzzy to me and I'm still sorting out my thoughts.

bseeger: But I can see that perhaps if that flag is in the omit, one could expect to not get the minimal containment triples, but I'm not sure fcrepo would do that.  (and it's possible that there's a bug here in fcrepo).

src/main/java/org/fcrepo/spec/testsuite/crud/Container.java
```diff
@@ -138,12 +139,23 @@ public void ldpcContainmentTriples() {
...
+                    getLocation(base), DIRECT_CONTAINER_BODY.replace("%membershipResource%", getLocation(container)));
+            directMember = createBasicContainer(getLocation(direct), "member");
+        } catch (SkipException e) { // DC not supported
+            directContainerTripleCheck = false;
+        }
 
+        // Only omit, not include, can guarantee the omission of specified triples.
         final Response resP = doGet(getLocation(container),
-                new Header("Prefer", "return=representation; include=\"http://www.w3.org/ns/ldp#PreferContainment\""));
+                new Header("Prefer", "return=representation; " +
```
gregjan: This line sets a Prefer headers that omits minimal and membership triples. Using "Omit" is the only way you can guarantee that a particular triple set is not present. Fedora doesn't seem to honor this currently, so I think we've identified a compliance gap w/LDP here.

src/main/java/org/fcrepo/spec/testsuite/crud/Container.java
gregjan: Yes, in the older version of this test did not check for "minimal" triples. That was a new check that I added when I realized the test wasn't quite comprehensive.

gregjan: It's a little complicated because these tests were using "include" before in order to test the absence of particular triples, but include doesn't guarantee absence, just inclusion. So I flipped these tests to use "omit".

src/main/java/org/fcrepo/spec/testsuite/crud/Container.java
bseeger: Hi @gregjan - I believe you're correct and have filed this bug accordingly: https://jira.duraspace.org/browse/FCREPO-3037

gregjan:
Okay, great! Glad that we identified a spec detail as well. I'm not sure that I can merge this, but let me know if there is more that I can do.

gregjan:
I'll wait to create my other PR until this merges, since it'll look a lot cleaner after I rebase..

PR #333, opened 2020-05-28
Some corrections

whikloj:
Nothing major and this doesn't resolve FCREPO-3325 but it needed to be fixed.

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/authz/WebACModes.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/authz/WebACModes.java
@@ -393,7 +393,7 @@ public class WebACModes extends AbstractAuthzTest {
                 + " WHERE { }";
 
         // Verify that non-admin user can update ACL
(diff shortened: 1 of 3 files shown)


src/main/java/org/fcrepo/spec/testsuite/authz/WebACModes.java
```diff
@@ -393,7 +393,7 @@ public void controlAllowedPUT() {
                 + " WHERE { }";
 
         // Verify that non-admin user can update ACL
-        doPatch(aclUri, new Headers(new Header("Content-Type", APPLICATION_SPARQL_UPDATE)), body, false);
+        doPutUnverified(aclUri, new Headers(new Header("Content-Type", APPLICATION_SPARQL_UPDATE)), body, false);
```
awoods, line 396: What does PUT with `Content-Type` of "application/sparql-update" mean? When I try to perform such a PUT on an "acl" or container, I get an error.

whikloj: Oops

PR #336, opened 2020-11-16
Allow PUT requests to fail if they express a constraint

### LINKED ISSUE #335: HttpPut test for 3.6.1-A doesn't allow for constraints.
Based on the [test description](https://github.com/fcrepo/Fedora-API-Test-Suite/blob/master/src/main/java/org/fcrepo/spec/testsuite/crud/HttpPut.java#L108-L112) this test should allow you to PUT **if** you don't try to alter server managed triples.

But the test creates a resource, then GETs the resource and tries to PUT the RDF back. This would by default include any server managed triples.

So we can change the test to allow for 4xx error messages but without viewing the specific implementations server managed triples it would be impossible to check that we have received any language about them.
> The server MUST provide a corresponding response body containing information about which statements could not be persisted.

whikloj:
Resolves #335

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/Constants.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/Constants.java
@@ -80,6 +80,9 @@ public class Constants {
 
     public static final String MEMENTO_DATETIME_HEADER = "Memento-Datetime";
 
+    public static final String LDP_NAMESPACE = "http://www.w3.org/ns/ldp#";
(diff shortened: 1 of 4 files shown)


whikloj:
I'll note that as I am "fixing" these tests we are just testing that Fedora returns a 4xx over and over again.

dbernstein:
@whikloj : I believe this fix is no longer needed with the introduction of the new  http://...ServerManaged omit type.

dbernstein:
This issue is no longer an problem with the introduction of the http://fedora.info/.../ServerManaged Prefer header option.

PR #342, opened 2021-01-15
Ensures that before and after comparisons are not including server ma…

dbernstein:
…naged timestamps that change when new mementos are created when 4.3.3.1-A and 4.3.3.1-C. This filtering is necessary because Fedora 6 will update the last modified date of resources when a new version is stamped. 

Resolves:
   * https://jira.lyrasis.org/browse/FCREPO-3596
   * https://jira.lyrasis.org/browse/FCREPO-3595

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpPost.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpPost.java
@@ -38,6 +38,9 @@ import org.testng.annotations.Test;
  */
 public class LdpcvHttpPost extends AbstractVersioningTest {
 
+    private static final String PREFER_MINIMAL_CONTAINER_HEADER = "return=representation; " +
+            "omit=\"http://fedora.info/definitions/v4/repository#ServerManaged\"";
+
     /**
      * 4.3 may disallow post
      */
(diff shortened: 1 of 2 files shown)


src/main/java/org/fcrepo/spec/testsuite/versioning/LdpcvHttpPost.java
```diff
@@ -38,6 +38,9 @@
  */
 public class LdpcvHttpPost extends AbstractVersioningTest {
 
+    private static final String PREFER_MINIMAL_CONTAINER_HEADER = "return=representation; " +
```
awoods, line 41: As you have seen, the difference between the original resource and the memento is the "last-modified" date. Such as:
```
< <http://localhost:8080/rest/testsuitecontainer01152021175807/2bd16409-e048-4bf8-b175-38c2889651d4> <http://fedora.info/definitions/v4/repository#lastModified> "2021-01-15T22:58:17.255455Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> .
---
> <http://localhost:8080/rest/testsuitecontainer01152021175807/2bd16409-e048-4bf8-b175-38c2889651d4> <http://fedora.info/definitions/v4/repository#lastModified> "2021-01-15T22:58:25.322368Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> .
```

I am a bit uncomfortable adding this Fedora-specific header into the API-Test-Suite... but the other options seem to be to post-process the RDF and ignore/remove certain values, like dates.

This works as is, but I will defer to the team if we are comfortable with this solution.

awoods (changes requested):

awoods:
In considering this further, filtering out the the 'lastmodified' triple may be the cleanest solution.

dbernstein:
Closing - this is resolved by https://github.com/fcrepo/fcrepo/pull/1866

PR #343, opened 2021-01-15
On 4.1.1-A-1 we now check if any mementos were created automatically …

dbernstein:
…on resource creation.  Only if they weren't do we check for a 406.

Resolves: https://jira.lyrasis.org/browse/FCREPO-3477

Test by running the testsuite and verifying that 4.1.1-A-1 passes.

### DIFF
--- a/src/main/java/org/fcrepo/spec/testsuite/versioning/LdprvHttpGet.java
+++ b/src/main/java/org/fcrepo/spec/testsuite/versioning/LdprvHttpGet.java
@@ -17,10 +17,11 @@
  */
 package org.fcrepo.spec.testsuite.versioning;
 
-import static java.time.format.DateTimeFormatter.ISO_INSTANT;
-import static java.time.format.DateTimeFormatter.RFC_1123_DATE_TIME;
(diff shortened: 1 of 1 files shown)


src/main/java/org/fcrepo/spec/testsuite/versioning/LdprvHttpGet.java
```diff
@@ -62,14 +62,27 @@ public void shouldReturn406WhenNoLdprm() {
...
         final URI timeGateUri = getTimeGateUri(response);
 
         //query timegate using Accept-Datetime
         final String isoDateString = "1970-01-01T00:00:00Z";
 
         final String rfc1123Date = convertToRfc1123DateTimeString(isoDateString);
 
-        doGetUnverified(timeGateUri.toString(), new Header("Accept-Datetime", rfc1123Date)).then().statusCode(406);
+        //if no mementos created verify 406 otherwise 302
+        doGetUnverified(timeGateUri.toString(), new Header("Accept-Datetime", rfc1123Date)).then()
```
awoods, line 84: I verified that this also passes when autoversioning is set to false, which is great.

awoods (approved):

PR #345, opened 2021-01-15
Fixes 3.10.2-D: the problem is Fedora 6 no longer accepts SMTs by def…

dbernstein:
…ault.

Resolves: https://jira.lyrasis.org/browse/FCREPO-3591

Test by running the test suite and verify 3.10.2-D passes.


whikloj (approved):
👍

